PR #4, opened 2024-07-09
Remove "mteb_version": "test" from dataset revisions

KennethEnevoldsen:
This pull request removes the "mteb_version": "test" field from the dataset revisions in multiple files. We instead assume that version is missing in these cases.

At some point we might consider outdating MTEB version (starting with "no version" files.)


orionw (approved):
LGTM

PR #5, opened 2024-07-09
remove task that does not exist within MTEB

KennethEnevoldsen:
I also attempted to figure out if it was renamed. I could not find that to be the case.


orionw (approved):
Wow good catch @KennethEnevoldsen! 

I assume these tasks are depreciated now but are old results? Does it hurt us to leave them in?

KennethEnevoldsen:
All of these are tasks that don't exist within MTEB. Couldn't even find then as deprecated tasks. My guess was that it was  internal tasks, which accidentally found its way to the leaderboard with another addition.

> Does it hurt us to leave them in?

Hmm at least MTEBResults will fail to load them as it tries to fetch the task data from MTEB.

Muennighoff:
Oh yes these were tasks that @staoxiao considered for CMTEB at some point but then decided to remove from CMTEB as he could not confirm the data source of them; Fine to remove I think 👍

PR #15, opened 2024-08-11
Add results

Muennighoff:


GritLM__GritLM-7B/13f00a0e36500c80ce12870ea513846a066004af/AFQMC.json
```diff
@@ -0,0 +1,43 @@
+{
+  "dataset_revision": "b44c3b011063adb25877c13823db83bb193913c4",
+  "evaluation_time": 17.481793880462646,
+  "kg_co2_emissions": 0.00635233445280417,
+  "mteb_version": "1.12.75",
+  "scores": {
+    "test": [
+      {
+        "cosine_pearson": NaN,
+        "cosine_spearman": NaN,
```
KennethEnevoldsen: what is going on here?

KennethEnevoldsen (commented):
Only a minor thing

GritLM__GritLM-7B/13f00a0e36500c80ce12870ea513846a066004af/AFQMC.json
Muennighoff: I'm not sure, there appear to be a few more NaNs; I think it's usually a bug in the dataset when there's empty rows

KennethEnevoldsen: It turns out that there are no labels on the "test" set. I have created a PR: https://github.com/embeddings-benchmark/mteb/pull/1153

Do we have contact to the CMTEB authors? This will likely influence their benchmark.

KennethEnevoldsen:
I moved the files into the `results` folder. This leads to multiple models appearing twice, which can cause problems, so we should probably merge the two folders for the models. As well as add a test in the future which prevents multiple folders of the same models.

Then I also reduced the size of files (flores) and added a script for doing it automatically.

Issue #24, opened 2024-09-05
Use local path.json


Samoed:
Currently, `path.json` is downloaded from the HF repository, which is inconvenient for local work. Why not use a local `path.json` instead of downloading it from the repository?

Samoed:
Ok, during debug figured out that paths file not downloading automatically

PR #25, opened 2024-09-05
fix models path

Samoed:
Since #18, you can't load results for models that are in the `author__model` folder format because the script only looks for folders that match the model name. For example:

```python
from datasets import load_dataset
ds = load_dataset("mteb/results", "e5-small-v2")
```

This would result in an error because there is only an `intfloat__e5-small-v2` folder. However, models without a revision still work since their folder names haven't changed.

I've also added a test to load all model results. I'm not sure how long this will take to run on CI, but locally, with a hard-coded local JSON, it took 2 minutes. Currently, it is failing due to errors like the one mentioned above.

@KennethEnevoldsen, could you please take a look at this?


Samoed:
Interesting. Locally I can run `make test`, but CI fails

KennethEnevoldsen:
This sounds like a change that should be fixed within `mteb`'s load function. Will add a fix there and the return to this issue.

Samoed:
I don't think so. Because this repo don't use mteb for work with data and source of problem wrong `paths.json`

__init__.py
KennethEnevoldsen, line 1: This looks like it should be deleted

results.py
KennethEnevoldsen, line 1: I am unsure why the tests passed in the linked PR but not here?

tests/test_load_results.py
```diff
@@ -24,3 +29,10 @@ def test_load_results():
     known_revision = "bf3bf13ab40c3157080a7ab344c831b9ad18b5eb"
     assert known_model in results
     assert known_revision in results[known_model]
+
+
+@pytest.mark.parametrize("model", MODELS)
+def test_load_results_from_datasets(model):
+    """Ensures that all models can be imported from dataset"""
+    path = Path(__file__).parent.parent / "results.py"
+    ds = load_dataset(str(path.absolute()), model, trust_remote_code=True)
```
KennethEnevoldsen: Why is this required? Does the leaderboard load it using datasets?

tests/test_load_results.py
```diff
@@ -1,6 +1,11 @@
 import os
 from pathlib import Path
 
+import pytest
+from datasets import load_dataset
+# for CI this should be `results.results`, for local just `results`
+from results.results import MODELS
```
KennethEnevoldsen: There shouldn't be a difference between the two. How do you run tests locally?

KennethEnevoldsen (changes requested):

tests/test_load_results.py
Samoed: Yes https://github.com/embeddings-benchmark/leaderboard/blob/d443c367474648aec8f145a73e0d7df5304f0a94/refresh.py#L245-L251

tests/test_load_results.py
Samoed: I run test with `make teset`

results.py
Samoed: Because previously test checked only folder structure and that mteb can import them, but not loading with load_dataset

KennethEnevoldsen:
@Samoed if you resolve a comment, please resolve - it makes it easier for me to see what you have fixed.

Also seems like there is still unresolved files

results.py
```diff
@@ -305,7 +321,7 @@ def _info(self):
 
     def _split_generators(self, dl_manager):
         path_file = dl_manager.download_and_extract(URL)
-        # Local debugging:
+        # Change path for local development
```
KennethEnevoldsen:
```suggestion
```

Samoed:
Now I resolved only `__init__` comment and was waiting for tests, but they didn't run. Other comments needed your review

KennethEnevoldsen:
> Other comments needed your review

Took a look over it again. Are you referring to:

> I don't think so. Because this repo don't use mteb for work with data and source of problem wrong paths.json

Are you saying that the leaderboard repo CI, assumes something about the file structure on this repo and that is what is causing the issue? 

The current file format is def. the one we want to stick with. So we should update the leaderboard code to match if that is the case.

Samoed:
> Are you saying that the leaderboard repo CI, assumes something about the file structure on this repo and that is what is causing the issue?

A bit yes, now in paths.json can be two versions of model result files
1. before adding author to model_name:
https://github.com/embeddings-benchmark/results/blob/45604e5b10b383a06d8725125e485273211da8e1/paths.json#L592-L596
2. and after that:
https://github.com/embeddings-benchmark/results/blob/45604e5b10b383a06d8725125e485273211da8e1/paths.json#L305-L309

And results repo looks only for models that are [defined in split ](https://github.com/embeddings-benchmark/results/blob/45604e5b10b383a06d8725125e485273211da8e1/results.py#L94) and now in repo for bge-m3 result will be files without revision and others with revision can't be imported. My PR merges these results by finding model_name after author name, so new format results can be used with old format

tests/test_load_results.py
```diff
@@ -24,3 +28,10 @@ def test_load_results():
     known_revision = "bf3bf13ab40c3157080a7ab344c831b9ad18b5eb"
     assert known_model in results
     assert known_revision in results[known_model]
+
+
+@pytest.mark.parametrize("model", MODELS)
+def test_load_results_from_datasets(model):
```
KennethEnevoldsen: This test does not belong here (this is for testing integration with MTEB's load_results function. Move it to another script

KennethEnevoldsen:
Hmm right, getting a better hang of the code here - the integration with mteb is quite poor here, which leads to a few issues.

So ideal situation:
```
from datasets import load_dataset
ds = load_dataset("mteb/results", "intfloat/e5-small-v2") # using just e5-small-v2 can lead to multiple matches
```

and that results should be compatible with

```
results = mteb.load_results(models=["intfloat/e5-small-v2"])
```

For splits etc. this is a duplicate of information already within `mteb`. A benchmark should only be specified once (and that is within the package). Split etc. should all be derived from `mteb`.

I don't see any reason why we should filter it within this repository (I might be missing something).

--- 

That is the ideal case. We might need to make some shortcuts here.

Samoed:
I understand your idea and agree that it should be implemented this way. However, it will require more time to integrate the old results into the new format, and some refactoring of the leaderboard will be needed

Samoed:
Now tests fails becase of https://github.com/huggingface/datasets/issues/7141

KennethEnevoldsen (approved):
> I understand your idea and agree that it should be implemented this way. However, it will require more time to integrate the old results into the new format, and some refactoring of the leaderboard will be needed

Thanks. Yep def. required more changes - I think this looks reasonable as is. Though we need to resolve the failing tests 

 @orionw since you are more familiar with the leaderboard integration, can I ask you to review this PR as well?

results.py
```diff
@@ -55,7 +56,6 @@
     "DanskBERT",
     "FollowIR-7B",
     "GritLM-7B",
-    "GritLM-7B-noinstruct",
```
orionw, line 58: What happened to this model?

results.py
```diff
@@ -193,14 +193,19 @@
     "sentence-t5-large",
     "sentence-t5-xl",
     "sentence-t5-xxl",
-    "sentence-transformers__LaBSE",
-    "sentence-transformers__all-MiniLM-L12-v2",
```
orionw, line 197: I see this model carried over -- did we just have all the other ones already? Like all-MiniLM-L6-v2?

orionw (commented):
Thanks for this PR @Samoed!! There was definitely some problem with it before that you fixed, thanks! I may be having a hard time following this PR, but it seems like we just make it so that any shortened model name works also so we don't miss any/have duplicates.

If so, did we make sure we kept all the existing results? I see the GritLM no instruction one got deleted for example.

That's my only concern, otherwise LGTM.

results.py
Samoed: There no results for this model in results folder

results.py
Samoed: Yes, all other models are already here

results.py
KennethEnevoldsen: Did we add these back in -why was this resolved?

results.py
KennethEnevoldsen: Did we add these back in -why was this resolved?

results.py
```diff
@@ -1,6 +1,7 @@
 """MTEB Results"""
 
 import json
+import os
```
KennethEnevoldsen: is this used?

KennethEnevoldsen:
@Samoed we still have to resolve the failing test before we can merge this in. Also please respond to the comments.

You also write:

> Currently, it is failing due to errors like the one mentioned above.

Which, when I read it along with the failing tests it seems like there is only partly a solution here. In which case please describe what is stopping it from becoming a finished solution.

results.py
Samoed: These models was already there

results.py
Samoed: I believe these results were deleted in #18 because they were present in the commit before that

Samoed:
I've added a test to load all model results by splits, but the results take path.json from the HF repository, where the paths are broken. To make this test pass, it needs to be fixed in the HF repository.

TIL: I wrote replies to all your comments, but I didn't know I needed to press the submit [button in the files tab](https://github.com/orgs/community/discussions/10369) to make them visible for you

results.py
Samoed:
```suggestion
```

results.py
KennethEnevoldsen: Ahh right, yes I did remove that. We should probably figure out a good approach to add results for experiment runs to the leaderboard (added an issue here: https://github.com/embeddings-benchmark/mteb/issues/1211)

KennethEnevoldsen: It was removed due to it conflicting with the results of gritLM with instructions (their model meta makes it seem like they are the same model when loaded from `mteb.load_results(...)`

results.py
KennethEnevoldsen: Right, but now we use the outdated version instead of the latest version of the results (which should be the same). I also believe that these new folders contain more results. It is not a big issue for now as we have plans to update the leaderboard, but it is worth noting the issues.

Samoed:
Currently, there is one model that won't be fixed by my PR: `text-embedding-preview-0409-768`. At the moment, there are two folders with the same model name, but the `author_name` is separated by a `.`. I think this should be standardized, but I'm not sure which model this corresponds to.
1. https://github.com/embeddings-benchmark/results/tree/main/results/google-gecko.text-embedding-preview-0409/no_revision_available
2. https://github.com/embeddings-benchmark/results/tree/main/results/google-gecko-256.text-embedding-preview-0409/no_revision_available

KennethEnevoldsen:
> I've added a test to load all model results by splits, but the results take path.json from the HF repository, where the paths are broken. To make this test pass, it needs to be fixed in the HF repository.

Thanks for the clarification should have noted that. Seems like a test that will never fail when we want it to (when we introduce an error it will use the current working version) and when that error is added the next (working) PR will fail. Shouldn't it just reference the path.json within the PR?

> TIL: I wrote replies to all your comments, but I didn't know I needed to press the submit [button in the files tab](https://github.com/orgs/community/discussions/10369) to make them visible for you

Ahh we have all been there :) 

re names, we should use '__' to separate the author from the model names. For now, let us treat 256 as a part of the authors (though my guess is that it refers to the embed size) so:

google-gecko-256.text-embedding-preview-0409/no_revision_available -> google-gecko-256__text-embedding-preview-0409/no_revision_available

Re: `text-embedding-preview-0409-768`

I can't seem to find that folder in the results repo?

Samoed:
> I can't seem to find that folder in the results repo?

I can't find it either, but I think it might be google-gecko.text-embedding-preview-0409. However, I'm not sure. Found these models in leaderboard config https://github.com/embeddings-benchmark/leaderboard/blob/12fdafc4eb41fb4e4dddfc6a6a202f13e23b734a/model_meta.yaml#L814-L831

> Shouldn't it just reference the path.json within the PR?

I thought the same regarding #24, but I believe that when HF tries to download the repository, it will only download results.py and README, not the paths file. So, the paths file is being downloaded by the results file at runtime, and I'm not sure how to improve this.

KennethEnevoldsen:
Let us delete `text-embedding-preview-0409-768` for now. The results are not there and the two on the leaderboard fall   under different names.

> I thought the same regarding https://github.com/embeddings-benchmark/results/issues/24, but I believe that when HF tries to download the repository, it will only download results.py and README, not the paths file. So, the paths file is being downloaded by the results file at runtime, and I'm not sure how to improve this.

hmm not sure either the code:

`ds = load_dataset(str(path.absolute()), model, trust_remote_code=True)`

does seem to follow the [docs](https://huggingface.co/docs/datasets/v2.21.0/en/loading#legacy-local-loading-script). It does say legacy though. If we can't get it to work we should just remove the test

Samoed:
> Let us delete `text-embedding-preview-0409-768` for now. The results are not there and the two on the leaderboard fall under different names.
> 
> > I thought the same regarding #24, but I believe that when HF tries to download the repository, it will only download results.py and README, not the paths file. So, the paths file is being downloaded by the results file at runtime, and I'm not sure how to improve this.
> 
> hmm not sure either the code:
> 
> `ds = load_dataset(str(path.absolute()), model, trust_remote_code=True)`
> 
> does seem to follow the [docs](https://huggingface.co/docs/datasets/v2.21.0/en/loading#legacy-local-loading-script). It does say legacy though. If we can't get it to work we should just remove the test

Yes, this will work for local development if full repo is provided, but when using `load_dataset("mteb/results")` code will give error, because it won't have results file

orionw:
Thanks for the response @Samoed, I understand now.

>  google-gecko.text-embedding-preview-0409

I have been confused by the difference between Google Gecko and this [text-embedding-004](https://cloud.google.com/vertex-ai/generative-ai/docs/model-reference/text-embeddings-api) model on the Arena. Google's naming conventions make this very hard to understand.

KennethEnevoldsen:
Thanks @Samoed for taking the time on this - I will merge it in now to allow the other PRs to be resolved as well

results.py
Muennighoff: Should we add this one back in? (Preliminary PR https://github.com/embeddings-benchmark/results/pull/28)

PR #28, opened 2024-09-10
re-add GritLM-7B-noinstruct

Muennighoff:
It seems like the `GritLM__GritLM-7B-noinstruct` results present here https://github.com/embeddings-benchmark/results/tree/319e81d64298f3bdc5f09c8b6ff91f5015b3664c/results/GritLM-7B-noinstruct are no longer present here https://github.com/embeddings-benchmark/results/tree/9a79f7e07542ad2f5cb47490fa1e5ac2ba57d7a8/results; probably they were accidentally deleted so this PR adds them back in & also merges the separate GritLM folders; Currently the regular GritLM-7B is missing from the [RAR-b leaderboard](https://huggingface.co/spaces/mteb/leaderboard) - this PR should fix it.


Samoed:
We've discussed why it was removed https://github.com/embeddings-benchmark/results/pull/25#discussion_r1750847260

Muennighoff:
I see; I guess we should add it back in but maybe make sure model_meta is different then?

results/GritLM__GritLM-7B-noinstruct/13f00a0e36500c80ce12870ea513846a066004af/model_meta.json
```diff
@@ -0,0 +1 @@
+{"name": "GritLM/GritLM-7B", "revision": "13f00a0e36500c80ce12870ea513846a066004af", "release_date": "2024-02-15", "languages": ["eng_Latn", "fra_Latn", "deu_Latn", "ita_Latn", "spa_Latn"], "n_parameters": null, "memory_usage": null, "max_tokens": null, "embed_dim": null, "license": null, "open_source": true, "similarity_fn_name": null, "framework": [], "loader": "gritlm_loader"}
```
KennethEnevoldsen:
```suggestion
{"name": "GritLM/GritLM-7B-no-instruct", "revision": "13f00a0e36500c80ce12870ea513846a066004af", "release_date": "2024-02-15", "languages": ["eng_Latn", "fra_Latn", "deu_Latn", "ita_Latn", "spa_Latn"], "n_parameters": null, "memory_usage": null, "max_tokens": null, "embed_dim": null, "license": null, "open_source": true, "similarity_fn_name": null, "framework": [], "loader": "gritlm_loader"}
```

KennethEnevoldsen: This is probably the quick fix, but we are discussing a more stable solution [here](https://github.com/embeddings-benchmark/mteb/issues/1211).

KennethEnevoldsen (commented):
With the name fix, I think we can add it in. Tests also still fail

Muennighoff:
Test is failing; I think it is because it is looking for `where exists = PosixPath('/home/runner/work/results/results/results/GritLM__GritLM-7B/no_revision_available/model_meta.json').exists`? Is the model_meta required even for `no_revision_available` folders? I think I can just move the results from that folder to the `13f00a0e36500c80ce12870ea513846a066004af` folder as I doubt the model was different

Samoed:
I think that you should update test fixture https://github.com/embeddings-benchmark/results/blob/main/tests/test_correct_folder_structure.py#L36-L248

Samoed:
I think that you can change fixture for test, because no_revision_available this is old result and GritLM versions are already presented here https://github.com/embeddings-benchmark/results/blob/b7f3700c5a62a4546b8045b9e70e882f3c70c48d/tests/test_correct_folder_structure.py#L36 or merge folders if results are the same or have different datasets

Muennighoff:
Makes sense, thanks! 

It seems like there is some difference in FollowIR results e.g. https://github.com/embeddings-benchmark/results/blob/standardizegrit/results/GritLM__GritLM-7B/no_revision_available/Robust04InstructionRetrieval.json vs https://github.com/embeddings-benchmark/results/blob/standardizegrit/results/GritLM__GritLM-7B/13f00a0e36500c80ce12870ea513846a066004af/Robust04InstructionRetrieval.json & same for others. I think it is because I reran all MTEB tasks including them with GritLM.
Should we update to the results from rerunning (in the folder with a commit id) @orionw? It seems like the FollowIR scores are higher after rerunning for all three tasks; Is it because of a difference in prompting?

orionw:
It could be, perhaps I did something wrong to the formatting.  As long as the instructions were correctly passed in and it looks like it would be unless there was a [prompt conflict](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/gritlm_models.py#L28) then it should be fine. 

I'm going to take a closer look at the instruction setup next week so I'm good with keeping the newer results!

PR #32, opened 2024-09-13
add split in results export

Samoed:
Description: https://github.com/embeddings-benchmark/leaderboard/issues/30
Related to: https://github.com/embeddings-benchmark/leaderboard/pull/33

@Muennighoff


Samoed:
@KennethEnevoldsen This should also be merged for the leaderboards to work correctly.

KennethEnevoldsen (approved):

KennethEnevoldsen:
Thanks again @Samoed. We have btw. discussed an updated version of the leaderboard here:
https://github.com/embeddings-benchmark/mteb/discussions/674

I have started implementing the shell of the idea. It will likely be a complete refactor e.g. since I don't want benchmarks to be both defined in the leaderboard and within MTEB (it should just be defined in MTEB and then automatically be added to the leaderboard). The hope is also that this will notably simplify the leaderboard code as much of the complexity will be implemented within MTEB.

Once I have an approximate outline I would love to get your input on it as well.

PR #46, opened 2024-11-11
add results for keeeeenw/MicroLlama-text-embedding

keeeeenw:
Hello,

I would like to add the results for https://huggingface.co/keeeeenw/MicroLlama-text-embedding

I ran the following code to get the results. Please let me know if I need to add more evaluation tasks for it to show up on the leaderboard. My model is for English language only.
```
import mteb
model = mteb.get_model("keeeeenw/MicroLlama-text-embedding")
tasks = mteb.get_benchmark("MTEB(eng, beta)")
evaluation = mteb.MTEB(tasks=tasks)
# the default 64 will get OOM on my Nvidia 4090
evaluation.run(model, output_folder="results", encode_kwargs={'batch_size':16})
```

Thanks!
Ken


Samoed:
Thank you for running your model! However, I don't think the results are accurate because you're using an instruct model, but `SentenceTransformer` won’t apply the prompts directly. To properly handle prompts, you’ll need to add your model to the MTEB models, similar to the approach shown [here](https://github.com/embeddings-benchmark/mteb/blob/fccf034bd78d74917f9d8fb6053e473fb03e86d8/mteb/models/e5_instruct.py#L19-L44).

keeeeenw:
Thanks for the feedback! When you say the results are not accurate, are they unusual low or high or just appear to be random? I’m very new to the embedding model-related projects so if the results are low, it is understandable and I will try to improve it. The base model I’m using is not instruction-tuned, and I used the normal sentence transformer training procedure (https://sbert.net/docs/sentence_transformer/training_overview.html).

results/keeeeenw__MicroLlama-text-embedding/no_revision_available/AmazonCounterfactualClassification.json
KennethEnevoldsen, line 1: It seems like you are also the creator of the model. In which case I would probably add a revision to the files both in the path and model_meta.json:

results/keeeeenw__MicroLlama-text-embedding/no_revision_available/AmazonCounterfactualClassification.json

You might also consider filling out the model_meta.json. (if you add the model to the [sentence transformers model registry](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/sentence_transformers_models.py) this will get filled out automatically.

KennethEnevoldsen (approved):

Samoed:
I think that if you try to include prompt results can be higher, because your base model is LLM

KennethEnevoldsen:
> I think that if you try to include prompt results can be higher, because your base model is LLM

Wouldn't that require training the model with prompts in mind?

Samoed:
Probably yes, so I could be wrong.

keeeeenw:
> Probably yes, so I could be wrong.

Yes, you will need to instruct tune the model by training it with a set of prompts. My base pretrained model performs the basic sequence-to-sequence next word prediction, and I modified it for text embedding, so theoretically it does not require any prompt. This is just my first attempt! I just want to get a baseline performance and iteration from there.

results/keeeeenw__MicroLlama-text-embedding/no_revision_available/AmazonCounterfactualClassification.json
keeeeenw: Sounds good! 
Can I generate any random hash such as "8b3219a92973c328a8e22fadcfa821b5dc75636a" for all_MiniLM_L6_v2 as long as it is unique to my model?

Samoed: Hash is the revision of the model from huggingface

keeeeenw: Make sense! I will make the change!

keeeeenw: I modified the folder name and content of model_meta.json with the revision hash of my model. Please take a look. Thanks for your review and feedback!

KennethEnevoldsen:
Looks good! will merge it in

keeeeenw:
Hi folks,

This pull request has been merged for a while but I am still not seeing the evaluation results for my model showing up on https://huggingface.co/spaces/mteb/leaderboard

I have only evaluated on ```tasks = mteb.get_benchmark("MTEB(eng, beta)")``` tasks so should I use a particular filter on the leaderboard or evaluate on additional tasks?

Thank you and happy holidays!

KennethEnevoldsen:
It should appear on https://huggingface.co/spaces/mteb/leaderboard_2_demo (to be fully released in january)

as long as the model is also added registrered on `mteb`

keeeeenw:
I don't think my model is registered on mteb yet. I have only had the changes locally. Let me submit a pull request.

PR #49, opened 2024-11-15
fix: devide external results metrics by 100

### LINKED ISSUE #48: Scores are multiplied by 100 by default in old results
Hey @Samoed! We accidentally used scores multiplied by 100 from the old leaderboard.
Can you fetch them and divide them by 100 as all other scores are in the benchmark?
Also can we add some safeguards against introducing duplicate results?

Samoed:
Closes #48 @x-tabdeveloping


x-tabdeveloping (approved):
Looks good, thanks!

PR #51, opened 2024-11-16
update results and path.json

small-starriest:
Update models by giving them instructions, and add results of datasets to relevant models.


small-starriest:
Sorry to bother you multiple times. I want to konw what is the difference between the external folder  and the no_revision-available folder in the model folder?

PR #56, opened 2024-11-26
fix: Enforce revision ID and model names for future contributions

KennethEnevoldsen:
Added tests to:
- Enforce revision ID for new models
- Enforce consistent model names for new models

Added this change mainly due to duplicates models results and due to recent addition not including a model revision (which I we should probably discourage).

This fixes model results for:
- voyage

(needed for the leaderboard)

PR #57, opened 2024-11-26
fix: remove duplicate models

KennethEnevoldsen:
- Added minor fixes from #56
- removed all vectoriseai/* models (e.g. vectoriseai/gte-large)

They seems to be duplicates of the original model.


Samoed:
You can add this username to https://github.com/embeddings-benchmark/results/blob/af028247d03c2d90539cbad9b024c123917f4c20/load_external.py#L209 to avoid having to add it again in the future.

KennethEnevoldsen:
Hmm but do we plan to download external results again?

I would argue that it is a one-time thing (since we have too many datasets now that the model card has become too large, >10 mb)

This also makes it easier to review incoming models

isaac-chung (approved):
Basically checked the filenames against [this list](https://github.com/embeddings-benchmark/results/pull/56#discussion_r1858829249).

PR #58, opened 2024-12-02
adding omnis results on STS12

gp-to:
Adding results of a test-embedding, which I titled omniembedding, on the STS12 dataset.


gp-to:
At first, I had some misunderstandings about how to include the revision_id, but it was all fixed in the last commit.

KennethEnevoldsen:
Hi @perticarari thanks for the PR :) We are working on a new version of the leaderboard  (v2), which should give a better overview. 

In that process, we are cleaning up the models, where users of the benchmark would not be able to interpret the results. 

I believe this might be such a case. A user would be unable to use the model or figure out how it got the results, therefore it does not add a lot of information to the leaderboard. 

Our decision has yet to be finalized so if you have strong points I would be more than happy to consider them - we want the benchmark to be as useable as possible

gp-to:
Hi, @KennethEnevoldsen 

I didn't quite understand what the issue is. What should I add / remove to get this PR approved? Is it related to the model card or model description?

x-tabdeveloping:
The canonically accepted way to do this would probably be to first implement your model along with detailed metadata in the `mteb` software package and then running your model that way.
Your PR is currently missing a lot of details including references to your model on HF Hub and model details.
We are also missing information on how you implemented your evaluations.
Please submit a PR to https://github.com/embeddings-benchmark/mteb including a model implementation along with metadata, then run your model with MTEB, then submit a PR here.

KennethEnevoldsen:
I will close this for now. Feel free to keep the discussion going, but currently it seems like this discussion has died out

PR #59, opened 2024-12-02
Add Gemma-Embeddings-v0.8 Retrieval Results

nicholasmonath:
Add Gofer-Embeddings-v0.8 results on the retrieval subset of tasks in MTEB.


KennethEnevoldsen:
Thanks for the PR @nicholasmonath it seems like multiple fields such as the MTEB version are not specified as shown by the tests. It also seems like the model_meta is not filled out.

Samoed:
Could you provide the script you used to run MTEB? It seems a bit unusual that the original results didn’t include the MTEB version and evaluation time

nicholasmonath:
Hi @KennethEnevoldsen and @Samoed. Thank you for your comments. We wrote a sanitizer to remove sensitive info like timings, but we realized that our sanitizer was overly sensitive and removed even necessary fields. We are working on updating the pull request. We will add back the MTEB version that we used (we noticed that it is actually an older version 1.0.3) and model_meta. However, we are still required to avoid evaluation time due to the sensitivity of the infrastructure that we use.

KennethEnevoldsen:
Excluding runtime and co2 emissions is fine, however, 1.0.3 is quite an old version. I would strongly recommend running it on the latest version of mteb. The scores should be approximately the same (minor differences as the seed changed in older version of the code along with code changes). We also standardize the result format in later versions of MTEB. If your model is prompt-based, newer versions of the benchmark allow you to integrate that as well.

nicholasmonath:
Hi @KennethEnevoldsen, thank you for comments and time reviewing this PR. We have now updated our MTEB version to 1.21.7. The latest files now have this version and we have only sanitized the evaluation time. Please let us know if you have any questions or concerns.

results/google__Gemma-Embeddings-v0.8/d6813d20532a97ea8e30fc285397d5105316511f/ArguAna.json
```diff
@@ -0,0 +1,158 @@
...
+        "nauc_mrr_at_1000_diff1": 0.411898,
+        "main_score": 0.83828,
+        "hf_subset": "default",
+        "languages": [
+          "eng-Latn"
+        ]
+      }
+    ]
+  },
+  "evaluation_time": -1,
```
KennethEnevoldsen: evaluation_time can't be -1. It should be null

results/google__Gemma-Embeddings-v0.8/d6813d20532a97ea8e30fc285397d5105316511f/model_meta.json
```diff
@@ -0,0 +1,24 @@
...
+  "public_training_data": null,
+  "public_training_code": null,
+  "framework": [],
+  "reference": null,
+  "similarity_fn_name": null,
+  "use_instructions": null,
+  "zero_shot_benchmarks": null,
+  "adapted_from": null,
+  "superseded_by": null,
+  "loader": null
```
KennethEnevoldsen:
```suggestion
  "languages": [
    "eng-Latn"
  ],
  "n_parameters": 9242164736,
  "memory_usage": null,
  "max_tokens": 8192,
  "embed_dim": 3584,
  "license": "gemma",
  "open_weights": False,
  "public_training_data": True,
  "public_training_code": False,
  "framework": ["API"],
  "reference": "https://huggingface.co/google/Gemma-Embeddings-v0.8",
  "similarity_fn_name": "cosine",
  "use_instructions": null,
  "adapted_from": "google/gemma-2-9b",
  "superseded_by": null,
  "loader": null
```

KennethEnevoldsen: Does the model use instruction?

KennethEnevoldsen (commented):
Thanks for the update - there is a few issues remaining

results/google__Gemma-Embeddings-v0.8/d6813d20532a97ea8e30fc285397d5105316511f/ArguAna.json
nicholasmonath: Thank you! I believe I have fixed this now.

nicholasmonath:
Thank you so much @KennethEnevoldsen! Sorry about those remaining issues. I believe I have resolved them all now. Please do let me know if there is anything else.

KennethEnevoldsen (approved):

nicholasmonath:
Hi @KennethEnevoldsen, thank you again for all of your help with this pull request. I wanted to check in about when the results would appear on the leaderboard? I thought that they might appear after the update today, but I don't see them added? Please let me know if there is anything more from my side that you need. 

Thanks very much!

KennethEnevoldsen:
For them to appear on the current leaderboard you will have to updatee paths.json (see snippet results.py) If adding a new models also add their names to results.py. 

(we are close to having a new leaderboard ready where this will no longer be necessary)

nicholasmonath:
Thank you for your quick reply and the information, @KennethEnevoldsen!  I have updated paths.json here: https://github.com/embeddings-benchmark/results/pull/69 

Note, that it looked like the MODELS in results.py are automatically pulled from this line: https://github.com/embeddings-benchmark/results/blob/main/results.py#L295 and so I did not modify this file.

PR #61, opened 2024-12-04
Add voyage-code-3 results

Liuhong99:
Add voyage-code-3 COIR results

Add `voyageai__voyage-code-3` in `paths.json`


paths.json
```diff
@@ -14696,6 +14696,18 @@
     "results/voyageai__voyage-3/no_revision_available/LegalQuAD.json",
     "results/voyageai__voyage-3/no_revision_available/XPQARetrieval.json"
   ],
+  "voyageai__voyage-code-3": [
```
KennethEnevoldsen, line 14699: Should we create CI to auto-update this?

KennethEnevoldsen (approved):
This looks very reasonable

Liuhong99:
Also PR in https://github.com/embeddings-benchmark/leaderboard/pull/58. Thanks!

Samoed:
To correctly add results, you need to generate `path.json` by running [`[results.py](https://github.com/embeddings-benchmark/results/blob/main/results.py)`](https://github.com/embeddings-benchmark/results/blob/main/results.py). Then, in the leaderboard repository, recreate `EXTERNAL_RESULTS.json` by deleting it and running `refresh.py`. Alternatively, you can wait for the release of Leaderboard 2.0

paths.json
Samoed: I don't think that we need, because soon will be `leaderboard2.0`

Liuhong99:
Please let me know how I can help to merge the results. Ideally we hope it could be on the LB before leaderboard 2.0 is ready. (btw, congratulations on the upcoming LB 2.0 release---This is huge for the community!)

paths.json
KennethEnevoldsen: I thought this was also used for the HF dataset?

KennethEnevoldsen:
> LB 2.0 release---This is huge for the community!

Glad to hear! I'm looking forward to getting it online as well!

> Please let me know how I can help to merge the results

It seems like I was a bit fast during my first review. It appears to fail due to the revision (we no longer allow "no_revision_available"). It seems like you used the implementation on [MTEB](https://github.com/embeddings-benchmark/mteb/blob/d713525824ab94e6269dad6dab2911c9bf71296c/mteb/models/voyage_models.py#L209) then you can simply add "1". 

Btw. the implementation on `mteb` is our reference implementation, so if you want to add the model there then we would be happy to review a PR (if there are disagreements between submitted scores and our implementation then we will prefer ours to ensure reproducibility)

paths.json
Samoed: For HF dataset yes, but I'm not shure if it popular, because mteb get results by cloning repository, but if you want I can add CI

Liuhong99:
Thanks! @KennethEnevoldsen. The revision is now fixed. We used MTEB implementation.

KennethEnevoldsen:
@Liuhong99 I have made some suggestions to the model_meta will you update it - otherwise I think we are good to go

Liuhong99:
> @Liuhong99 I have made some suggestions to the model_meta will you update it - otherwise I think we are good to go

@KennethEnevoldsen might be a dumb question--where can I locate the suggestions to model_meta?
Below is the current model_meta:

`{"name": "voyageai/voyage-code-3", "revision": "1", "release_date": "2024-12-02", "languages": null, "n_parameters": null, "memory_usage": null, "max_tokens": 32000, "embed_dim": 1024, "license": null, "open_source": false, "similarity_fn_name": null, "framework": [], "loader": "VoyageWrapper"}`

KennethEnevoldsen:
> @KennethEnevoldsen might be a dumb question--where can I locate the suggestions to model_meta?
Below is the current model_meta:

Do you mean where to add the changes?

You can either change the json directly, but ideally, we would want these to be submitted to the MTEB repository along with the model "voyageai/voyage-code-3".

Liuhong99:
If the change refers to changing "no_revision_available" to "1", then it's already fixed and committed. Please let me know if there are other places that you suggest to change in model_meta. Thanks ❤️

KennethEnevoldsen:
Sorry seems like I misunderstood the question, the you can find the suggestion here (screenshot from earlier in the thread):

Screenshot 2024-12-10 at 16 13 51

Liuhong99:
Got it! Makes perfect sense. Updated model_meta.json

KennethEnevoldsen:
Wonderful - merged!

PR #65, opened 2024-12-09
feat: evaluate openai, cohere and google models in CUREv1

dbuades:
As a follow-up to https://github.com/embeddings-benchmark/results/pull/55, this PR evaluates the following models on CUREv1:

  - name: "openai/text-embedding-3-small"
  - name: "openai/text-embedding-3-large"
  - name: "Cohere/Cohere-embed-english-v3.0"
  - name: "Cohere/Cohere-embed-english-light-v3.0"
  - name: "Cohere/Cohere-embed-multilingual-v3.0"
  - name: "Cohere/Cohere-embed-multilingual-light-v3.0"
  - name: "google/text-embedding-005"
  - name: "google/text-multilingual-embedding-002"

These results were obtained using the code from the following two PRs: https://github.com/embeddings-benchmark/mteb/pull/1562 and https://github.com/embeddings-benchmark/mteb/pull/1564.

As a side note, we would like to run these models on the remaining tasks in the MTEB(Medical) benchmark. However, we initially held off due to API cost constraints. Do you have access to credits with these providers that we could use for this purpose? Alternatively, would it be possible for you to run them on your side?

## Checklist

- [x] Run tests locally to make sure nothing is broken using `make test`. 
- [x] Run the results files checker `make pre-push`.

PR #71, opened 2024-12-13
feat: evaluate `openai` models on the remaining `MTEB(Medical)` tasks

dbuades:
Following up on [this PR](https://github.com/embeddings-benchmark/results/pull/65), this PR adds results from evaluating `text-embedding-3-small` and `text-embedding-3-large` on the remaining tasks in the `MTEB(Medical)` benchmark.

As discussed [here](https://github.com/embeddings-benchmark/results/pull/65#issuecomment-2533194935), results from revision 1 are equivalent to those from revision 2. Therefore, we only evaluated tasks that were not previously run.

Thank you @Muennighoff for providing an API key with credits!

## Checklist


KennethEnevoldsen (approved):
Looks good!

I just noted that we track co2 for APIs, which seems a bit misleading. Not sure what the best solution is here (my guess would be that we just filter it out in the leaderboard)

dbuades:
@Muennighoff I like your suggestion of also running both models with reduced dimensions. I was thinking 768 dimensions would be a good number since it would put it in line with medium sized models, testing the compression accuracy of their Matryoshka training. What do you think?

Muennighoff:
> @Muennighoff I like your suggestion of also running both models with reduced dimensions. I was thinking 768 dimensions would be a good number since it would put it in line with medium sized models, testing the compression accuracy of their Matryoshka training. What do you think?

Feel free to run all of them if you want :)

dbuades:
> Looks good!
> 
> I just noted that we track co2 for APIs, which seems a bit misleading. Not sure what the best solution is here (my guess would be that we just filter it out in the leaderboard)

That was fast!

dbuades:
> > @Muennighoff I like your suggestion of also running both models with reduced dimensions. I was thinking 768 dimensions would be a good number since it would put it in line with medium sized models, testing the compression accuracy of their Matryoshka training. What do you think?
> 
> Feel free to run all of them if you want :)

Thanks! I'll run all the common dimensions then (256, 512, 768, 1024)

KennethEnevoldsen:
> That was fast!

caught me at boring conference talk ;)

KennethEnevoldsen:
Re running the dimensions: I just want to refer to https://github.com/embeddings-benchmark/mteb/issues/1211

dbuades:
> > That was fast!
> 
> caught me at boring conference talk ;)

By the way, regarding CodeCarbon, in addition to filtering it out in the leaderboard, we could also modify mteb.run() to check if the evaluated model is an instance of one of the API provider wrappers. If so, we could override the co2_tracker parameter to False. Somewhere around [here](https://github.com/clinia/mteb/blob/eecc9f1192bdf7cb63bd7151359629cae0b87fd3/mteb/evaluation/MTEB.py#L499).

KennethEnevoldsen:
> By the way, regarding CodeCarbon, in addition to filtering it out in the leaderboard, we could also modify mteb.run() to check if the evaluated model is an instance of one of the API provider wrappers

But I am not sure we know at that time (the encoder does not contain the metadata)

dbuades:
> Re running the dimensions: I just want to refer to [embeddings-benchmark/mteb#1211](https://github.com/embeddings-benchmark/mteb/issues/1211)

I hadn't seen this, thanks. I intended to simply create a PR in `mteb` doing something like:

```
text_embedding_3_small_512 = ModelMeta(
    name="text-embedding-3-small-512",
    revision="1",
    release_date="2024-01-25",
    languages=None,  # supported languages not specified
    loader=partial(
        OpenAIWrapper,
        model_name="text-embedding-3-small",
        tokenizer_name="cl100k_base",
        max_tokens=8192,
        embed_dim=512,
    ),
    max_tokens=8191,
    embed_dim=512,
    open_weights=False,
    n_parameters=None,
    memory_usage=None,
    license=None,
    reference="https://openai.com/index/new-embedding-models-and-api-updates/",
    similarity_fn_name="cosine",
    framework=["API"],
    use_instructions=False,
)

text_embedding_3_small_768 = ModelMeta(
    name="text-embedding-3-small-768",
    revision="1",
    release_date="2024-01-25",
    languages=None,  # supported languages not specified
    loader=partial(
        OpenAIWrapper,
        model_name="text-embedding-3-small",
        tokenizer_name="cl100k_base",
        max_tokens=8192,
        embed_dim=768,
    ),
    max_tokens=8191,
    embed_dim=768,
    open_weights=False,
    n_parameters=None,
    memory_usage=None,
    license=None,
    reference="https://openai.com/index/new-embedding-models-and-api-updates/",
    similarity_fn_name="cosine",
    framework=["API"],
    use_instructions=False,
)
```

but the experiments approach is more flexible.

KennethEnevoldsen:
I think this is perfectly file for running them, but I don't think we would accept the PR as we have been working on removing duplicates on the new leaderboard and this would add them again.

(would love to have them run though, then we have start experimenting with how to best display it)

dbuades:
> > By the way, regarding CodeCarbon, in addition to filtering it out in the leaderboard, we could also modify mteb.run() to check if the evaluated model is an instance of one of the API provider wrappers
> 
> But I am not sure we know at that time (the encoder does not contain the metadata)

I agree that the encoder does not contain the metadata, but I believe `mteb.run()` does since it receives the model as an argument. We could check the model meta and if the framework is `API`, then disable co2_tracker.

Maybe I'm not explaining myself clearly, I can open a small PR to illustrate it in the coming days.

dbuades:
> I think this is perfectly file for running them, but I don't think we would accept the PR as we have been working on removing duplicates on the new leaderboard and this would add them again.
> 
> (would love to have them run though, then we have start experimenting with how to best display it)

Fair enough! I'll run the evaluations for now and open a draft PR here (they would still show up on different folders) but not on `mteb`.

KennethEnevoldsen:
Would love a PR - should make it clear

dbuades:
> Would love a PR - should make it clear 

Perfect, I'll open one.

PR #76, opened 2024-12-19
Add RuMTEB evaluation scores for most english top MTEB models

AlexeyVatolin:
For the models `gte-Qwen2-7B-instruct`, `NV-Embed-v1`, `NV-Embed-v2`, `SFR-Embedding-2_R`, `gte-Qwen1.5-7B-instruct`, `gte-Qwen2-1.5B-instruct` no scorers for the `MIRACLRetrieval` task have been computed yet, they are in progress. I will add them separately as soon as they are finished. 

## Checklist

### Adding a model checklist

- I have added model implementation to `mteb/models/` [directory](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models). Instruction to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md)


Samoed:
Are you sure it's a good idea to add NV-Embed if we can't match the authors' implementation?

AlexeyVatolin:
@Samoed I'm not sure, but it's the best we have at the moment. If we can fix the NV-Embed model implementation and get the better results, I will update the scores on RuMTEB.

Samoed (approved):

PR #77, opened 2024-12-20
add giga-embedding-instruct model

ekolodin:
## Checklist

### Adding a model checklist

Ref # 

- I have added model implementation to `mteb/models/` [directory](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models). Instruction to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md)


Samoed:
Could you run `MiraclReranking` with the current MTEB version since its score format has changed? Also, could you add the implementation to the MTEB repository?

KennethEnevoldsen (changes requested):
Sorry for the delay on this

The result format is good.

~~Just to make you aware you will have to submit a model meta to the mteb repo for it to appear on the leaderboard.~~

~~The model meta file is also unfilled~~

Samoed:
@KennethEnevoldsen I was trying to add this model in https://github.com/embeddings-benchmark/mteb/pull/1741

KennethEnevoldsen:
Ah yes I see - I could have seen that from the reference

KennethEnevoldsen:
Re-reviewed this PR. Seems like the revision is missing. I don't think we can merge it without it. However @Samoed's implementation has been merged so it should be easy to rerun. 

Will close this PR  for now, but do open a new one

PR #80, opened 2024-12-21
fix: external download

Samoed:
## Checklist

It seems that some tasks are using data from the same repo and I used repo name as key to store results. Now I've changed it to task name. Also added renamed datasets (`NorwegianParliamentClassification`, `CMedQAv1-reranking`, `EightTagsClustering`, `PawsXParaphraseIdentification`).

ref https://github.com/embeddings-benchmark/mteb/issues/1571


KennethEnevoldsen:
Nice!!

PR #88, opened 2025-01-08
Add MTEB output for Google's text-embedding-005

chenblair:
Output is generated by in-house Google evaluation script forked off of original MTEB evaluation code

## Checklist


Muennighoff (approved):
Amazing! @KennethEnevoldsen is `external` the right folder? When naming the folder `no_revision_available` the tests failed for @chenblair. I had assumed that `1` would be the wrong folder to put it in since that one is only for results run with https://github.com/embeddings-benchmark/mteb/blob/25f4f618f1694d1155919c9771c551fa70b5049b/mteb/models/google_models.py#L157 while these results are probably run with the internal Google implementation. Or is `1` fine?

results/google__text-embedding-005/external/model_meta.json
```diff
@@ -0,0 +1,26 @@
...
+        "eng-Latn"
+    ],
+    "n_parameters": null,
+    "memory_usage": null,
+    "max_tokens": 2048.0,
+    "embed_dim": 768,
+    "license": null,
+    "open_weights": false,
+    "public_training_data": null,
+    "public_training_code": null,
```
KennethEnevoldsen:
```suggestion
    "public_training_data": null,
    "public_training_code": null,
```
```suggestion
    "public_training_data": false,
    "public_training_code": false,
```
assumed

results/google__text-embedding-005/external/model_meta.json
```diff
@@ -0,0 +1,26 @@
...
+    "framework": [
+        "API"
+    ],
+    "reference": null,
```
KennethEnevoldsen: No release reference?

results/google__text-embedding-005/external/model_meta.json
```diff
@@ -0,0 +1,26 @@
...
+    "similarity_fn_name": "cosine",
+    "use_instructions": true,
+    "training_datasets": null,
```
KennethEnevoldsen, line 22: If you know that the model is not trained on any of the dataset in mteb (including other benchmarks), we can add:

```suggestion
    "training_datasets": [],
```
In which case it will appear as a zero-shot otherwise it will state that the dataset is unknown in the v2 of the benchmark (this is the demo, i.e. not released yet):

400787265-73ff3832-6208-4e65-b117-7f1929dd2228

results/google__text-embedding-005/external/AmazonCounterfactualClassification.json
```diff
@@ -0,0 +1,17 @@
+{
+  "dataset_revision": "e8379541af4e31359cca9fbcf4b00f2671dba205",
+  "mteb_dataset_name": "AmazonCounterfactualClassification",
+  "mteb_version": "1.0.3.dev0",
```
KennethEnevoldsen, line 4: This is quite an old version of MTEB. In the future, we might outdate results from before v1.11.0 (due to file format and a few notable bug fixes).

I would recommend rerunning this using the latest version.

results/google__text-embedding-005/external/ArguAna.json
```diff
@@ -0,0 +1,38 @@
+{
+  "dataset_revision": null,
```
KennethEnevoldsen: Seems to be missing for multiple datasets.

KennethEnevoldsen:
> Amazing! @KennethEnevoldsen is external the right folder?

External should be fine, unless it is in fact run with our implementation then it is in fact `1`. If not I would recommend that they make a PR of their implementation (allows users to know how they should use the API).

> Output is generated by in-house Google evaluation script forked off of original MTEB evaluation code

Explains some of the point. Would rerun it using the latest version of the public code-base.  

It seems like you are already using the implemented wrapper (from the model_meta.json): 

https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/google_models.py

So I would assume that it should work out of the box.

results/google__text-embedding-005/external/model_meta.json
chenblair: The training datasets haven't been revealed, so I will keep it as "null" for now.

results/google__text-embedding-005/external/model_meta.json
chenblair: Thanks for the catch! Added a reference to the text embedding documentation.

results/google__text-embedding-005/external/AmazonCounterfactualClassification.json
chenblair: I will leave this as a TODO in a future eval since it require some effort to run the latest version with our infrastructure properly.

results/google__text-embedding-005/external/ArguAna.json
chenblair: Added!

chenblair:
Thanks for the detailed comments and discussion! I will look into making a PR of our current implementation, the metrics as is should be ready to merge.

KennethEnevoldsen:
Fixes look good, I have this in

chenblair:
@Muennighoff 

I'm not currently seeing the [text-embedding-005](https://cloud.google.com/vertex-ai/generative-ai/docs/embeddings/get-text-embeddings#latest_models) metrics on the main MTEB leaderboard (https://huggingface.co/spaces/mteb/leaderboard). Is there anything else I need to do to have this model be populated on the leaderboard?

Thanks!

Muennighoff:
I think the `paths.json` file hasn't been updated? You need to run this function from the source of the repo https://github.com/embeddings-benchmark/results/blob/c33da340e109a8755faeda270b17502d095aef5b/results.py#L298 I think? cc @KennethEnevoldsen 

The new leaderboard will no longer require that and I think it already shows up there (http://mteb-leaderboard-2-demo.hf.space/?benchmark_name=MTEB%28eng%2C+classic%29)

Samoed:
> I think the paths.json file hasn't been updated? You need to run this function from the source of the repo.

Also to be added to old leaderboard you need to update [model_meta.yaml](https://github.com/embeddings-benchmark/leaderboard/blob/main/model_meta.yaml) and run [refresh.py](https://github.com/embeddings-benchmark/leaderboard/blob/main/refresh.py)

PR #94, opened 2025-01-21
update results for  Alibaba-NLP/gte-modernbert-base

afalf:
## Checklist

- [ x ] Run tests locally to make sure nothing is broken using `make test`. 
- [ x ] Run the results files checker `make pre-push`.

### Adding a model checklist

- [ x ] I have added model implementation to `mteb/models/` [directory](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models). Instruction to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) in the following PR ____


x-tabdeveloping:
I believe the reason for the conflict is that #97 contains most of these. Can you try resolving the conflict and see if there is anything still missing from the repo? @afalf

Samoed:
Results shouldn’t be in the `external` directory because that’s reserved for results from Hugging Face model cards. Your results should be placed in the directory corresponding to the model revision.

PR #95, opened 2025-01-21
add missing models

Samoed:
## Checklist

In a previous PR, I accidentally removed some models, which caused the old leaderboard to stop updating. I've now added the old results back. https://github.com/embeddings-benchmark/leaderboard/pull/64#issuecomment-2603873749

@x-tabdeveloping @KennethEnevoldsen can you review PR to merge it?

CC @sam-hey


results/flaubert__flaubert_base_cased/no_revision_available/model_meta.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "flaubert/flaubert_base_cased",
```
x-tabdeveloping, line 2: Hmm should this be empty? Why add a model_meta.json then?

x-tabdeveloping (approved):
Looks rather reasonable to me!

results/flaubert__flaubert_base_cased/no_revision_available/model_meta.json
Samoed: I think I previously removed these models because the tests were failing due to the absence of a `model_meta.json`. I've added them now to make the tests pass.

x-tabdeveloping: Alright. Do we have metadata annotations for this in MTEB?

Samoed: No, but these models were in old leaderboard and needed to be there for updating it. I think this is last update of old leaderboard https://github.com/embeddings-benchmark/leaderboard/pull/64

Samoed: Or maybe we can just ignore this, since new leaderboard will be very soon. You can write your suggestion in https://github.com/embeddings-benchmark/leaderboard/pull/64

PR #97, opened 2025-01-22
fetch more external results

Samoed:
## Checklist


Samoed:
It seems that cde models wasn't loaded. I'll try to figured out why

x-tabdeveloping (approved):
Looks quite reasonable

PR #99, opened 2025-01-26
skip test

Samoed:
This test was used to check if all models could be downloaded using `datasets`, but now we get the data by cloning the repository, so this test is no longer needed.

## Checklist


Samoed:
@x-tabdeveloping @KennethEnevoldsen Can you merge this PR to speed up tests?

x-tabdeveloping (approved):

PR #107, opened 2025-02-04
Ensure external does not have a revision

KennethEnevoldsen:
External models had revision. This lead to duplicate scores (two results from the same model on the same revision).

I have replace the revision with "no_revision_available". I also think this more accurately reflects the reality (we don't know the version they used when they ran the model).

## Checklist


Samoed:
Can you update the load external script too?

KennethEnevoldsen:
I think we should avoid using the `load_external` (model submission should happen here). So I have actually deleted it.

KennethEnevoldsen:
Let me know if you agree 

I also added a few extra tests these should make the revision much more consistent in the future (they also disallow "external")

essentially you are allowed to use a sha1 rev id or an integer (1, 2, 3) in case of APIs

KennethEnevoldsen:
(@x-tabdeveloping just so that you see this)

Samoed:
Yes, I agree that with new leaderboard this script can be deleted

Samoed (approved):

PR #143, opened 2025-03-07
Add MTEB(multilingual,v1),(en, v2) and (code,v1) for  gemini-embedding-exp-03-07

Feiyang1:
## Checklist

### Adding a model checklist

- [X] I have added model implementation to `mteb/models/` [directory](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models). Instruction to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) in the following PR - https://github.com/embeddings-benchmark/mteb/pull/2279


Feiyang1:
@Muennighoff any idea why the test is failing?

Muennighoff:
Hmm not sure why

Feiyang1:
Still failing. I'll try creating a new PR with only multilingual results

Muennighoff:
I think the issue is that you are naming some things `mmteb` in your result field but it should be `mteb`

Muennighoff:
Did you manually rename these? Cause the codebase should not do that I think

PR #151, opened 2025-03-10
Add results for voyage model

KennethEnevoldsen:
Added result for voyage models for all benchmarks

Also added required upper bound for mteb (python<=3.12)

Added a script for running all voyage models (I think these would be nice to just quickly update some results). Let me know what you think.

## Checklist


scripts/run_models/results/voyageai__voyage-3/1/BornholmBitextMining.json
Samoed, line 1: I think we don't need this folder

Samoed (approved):

scripts/run_models/run_voyage_models.py
Samoed, line 1: Do we need to save this script?

KennethEnevoldsen: removed

PR #153, opened 2025-03-12
Add results for baseline models for all benchmarks  

KennethEnevoldsen:
Resolves:
https://github.com/embeddings-benchmark/mteb/issues/2282#issuecomment-2716317536

Added results for:
`sentence-transformers/static-similarity-mrl-multilingual-v1`

for all benchmarks. I also ran a bunch of other models with the intention of having all of them run for all benchmarks. However, some of the benchmarks are huge, so I removed a lot of those.

With this change, we should no longer accept benchmarks without a set of evaluated models.

## Checklist


Samoed (approved):
Great!

PR #159, opened 2025-03-21
Add more results BEIR-NL (2)

nikolay-banar:
More models for BEIR-NL

## Checklist

### Adding a model checklist


KennethEnevoldsen (changes requested):
Seems like you are missing revisions on all models as well as metadata. That is why the test fails.

Edit: I have also added the checklist back in. Please do not delete it.

nikolay-banar:
> Seems like you are missing revisions on all models as well as metadata. That is why the test fails.
> 
> Edit: I have also added the checklist back in. Please do not delete it.

I added a wrong model. Sorry for that. It should be fine now.

KennethEnevoldsen (approved):

PR #165, opened 2025-03-26
add Ops-MoA-Yuan-embedding-1.0 and Ops-MoA-Conan-embedding-v1 C-MTEB Retrieval 

ahxgw:
## Checklist

### Adding a model checklist

- [ ] I have added model implementation to `mteb/models/` [directory](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models). Instruction to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) in the following PR ____


results/OpenSearch-AI__Ops-MoA-Conan-embedding-v1/external/CmedqaRetrieval.json
Samoed, line 1: You shouldn't add your results in `external` folder. You should implement your model in MTEB

PR #180, opened 2025-04-18
MTEB (eng, v2) for RELLE embedding

24September:
## Checklist

### Adding a model checklist


Samoed (changes requested):
Can you integrate your model to the mteb? You can follow guide from PR description

24September:
We did PR last Friday, and integrated our model to the mteb as you requested., but seems it is not approved. Do we need PR again?

PR #201, opened 2025-05-20
Rename gemini-embedding-exp-03-07 to gemini-embedding-001

Feiyang1:
### Checklist

- [ ] I *solemnly swear* that for all results submitted I have not training on the dataset including the training set. If I have I have disclosed it clearly.


Feiyang1:
@Muennighoff Could we get it merged? Thanks!

KennethEnevoldsen (approved):

PR #210, opened 2025-05-31
Add script for computing results table for PRs

KennethEnevoldsen:
This add the script for computing the PR results table comment.

In another PR we could turn this into and action like `@gitbot compare intfloat/multilingual-e5-large myorg/my-new-model`

but this is a good start.


Samoed (approved):
Looks good. I think it would be easier to create a ci that would generate comment. I'll do it next week

KennethEnevoldsen:
Yeah this is just the first step. Should be easy enough to adapt.

PR #203, opened 2025-05-25
add cadet results

manveertamber:
### Checklist

- [x] My model has a model sheet, report or similar
- [x] My model has a reference implementation in [`mteb/models/`](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models) this can be as an API. Instruction on how to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) 
  - [ ] No, but there is an existing PR ___
- [x] The results submitted is obtained using the reference implementation
- [x] My model is available, either as a publicly accessible API or publicly on e.g., Huggingface
- [x] I *solemnly swear* that for all results submitted I have not training on the dataset including the training set. If I have I have disclosed it clearly.

PR #218, opened 2025-06-10
add cadet results

manveertamber:
Not sure why but the added files from https://github.com/embeddings-benchmark/results/pull/203 are missing. Submitting another PR.

### Checklist

  - [] No, but there is an existing PR ___


KennethEnevoldsen:
Hmm, not sure - but will merge this in again - sorry for the inconvenience

PR #220, opened 2025-06-11
init automate script

### LINKED ISSUE #196: Generate table with submited results
Create table with newly submitted results in comments to the PR. I think table can look like

**Model name**
| Task name | Results | Rank |
| - | - | - |
| Task name 1 | 0.5 | 3 |
| Task name 2 | 0.9 | 1 |

Samoed:
Closes https://github.com/embeddings-benchmark/results/issues/196

I've created CI action that would create tables for PR automatically. If there won't be any results for `multilingual-e5` or `gemini-embedding-001` then results of only new model will be added.

With each new commit CI will update comment with results.

### Checklist

- [ ] I *solemnly swear* that for all results submitted I have not on the evaluation dataset including training splits. If I have I have disclosed it clearly.


scripts/pr_results_comment.py
KennethEnevoldsen, line 1: Shouldn't we only have one CLI for this?

If so, I would add the CLI like this:

```
Arguments:
  --files file1.json file2.json # if not specified, if will just get the diff from main (code in my script)
  --output results.md          
  --models intfloat/e5-large-v2 # if not specified will simply use default models
```

This way we can call it both from an action, but also from the command line

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+    """Group files by model and extract task names."""
+    model_tasks = defaultdict(list)
+
+    for file_path in file_paths:
+        try:
+            model_name, task_name = extract_model_and_task_from_path(file_path)
+            model_tasks[model_name].append(task_name)
+            print(f"✓ Found: {model_name} -> {task_name}")
+        except ValueError as e:
+            print(f"⚠ Warning: Skipping {file_path}: {e}")
```
KennethEnevoldsen: use the logging library

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+            return model_name, task_name
+        else:
+            raise ValueError(f"Invalid path structure: {file_path}")
+    except ValueError:
+        raise ValueError(f"Path must contain 'results' directory: {file_path}")
+
+
+def group_files_by_model(file_paths: list[str]) -> dict[str, list[str]]:
+    """Group files by model and extract task names."""
+    model_tasks = defaultdict(list)
```
KennethEnevoldsen:
```suggestion
    model_tasks = defaultdict(set)
```

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+    for file_path in file_paths:
+        try:
+            model_name, task_name = extract_model_and_task_from_path(file_path)
+            model_tasks[model_name].append(task_name)
+            print(f"✓ Found: {model_name} -> {task_name}")
+        except ValueError as e:
+            print(f"⚠ Warning: Skipping {file_path}: {e}")
+            continue
+
+    # Remove duplicates and sort
```
KennethEnevoldsen:
```suggestion
```

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+    for model in model_tasks:
+        model_tasks[model] = sorted(list(set(model_tasks[model])))
+
+    return dict(model_tasks)
+
+
+def create_comparison_table(model: str, tasks: list[str]) -> pd.DataFrame:
+    """Create comparison table for given models and tasks."""
+    try:
```
KennethEnevoldsen: Why the try? Just raise the error

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+
+        # Calculate averages only for numeric columns
+        averages = {}
+        for model in model_names:
+            if model in df.columns:
+                numeric_values = pd.to_numeric(df[model], errors='coerce')
+                avg_value = numeric_values.mean()
+                averages[model] = avg_value if not pd.isna(avg_value) else None
+            else:
+                averages[model] = None
```
KennethEnevoldsen: You should know what columns are numeric, no loop here simply index.

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def create_comparison_table(model: str, tasks: list[str]) -> pd.DataFrame:
+    """Create comparison table for given models and tasks."""
+    try:
+        print(f"Loading results for model: {model}")
+        print(f"Tasks: {tasks}")
+
+        try:
+            results = mteb.load_results(models=[model]+REFERENCE_MODELS, tasks=tasks, download_latest=False)
```
KennethEnevoldsen: Does this fail? It shouldn't

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+        "--output",
+        type=Path,
+        required=True,
+        help="Output markdown file path (required)",
+    )
+    return parser
+
+
+def main():
+    """Main function."""
```
KennethEnevoldsen:
```suggestion
```

scripts/pr_results_comment.py
```diff
@@ -0,0 +1,279 @@
...
+    parser = create_argparse()
+    args = parser.parse_args()
```
KennethEnevoldsen: I don't know if there is a general recommendation on this but I prefer keeping parsing outside the main function:

```
 if __name__ == "__main__":
    parser = create_argparse()
     args = parser.parse_args()
     main(files=files, output=args.output)
```

This way, you prevent any manipulation of the args. You also state what you expect to get as input to main.

scripts/pr_results_comment.py
KennethEnevoldsen: Generally, there are a lot of try statements here that I believe could be removed. Try statements generally make the code hard to debug as the error is not raised as the time of the error (e.g. scope is lost when running it in a debugger)

KennethEnevoldsen (commented):
Looks good. 

A few things which we could add, but is not required to merge this (probably out of scope):
- Would love to also include the "max" score the each task
- Would love to be able trigger this command using something like @bot result --models intfloat/e5-large-v2

Samoed:
> Would love to be able trigger this command using something like @bot result --models intfloat/e5-large-v2

To use it like a bot, we need to host it somewhere. I can host it, but I'm not sure if that's the best option. We could make it trigger on each comment and check if there’s a command to run, but I’m not sure if that’s a good idea.

Another solution could be to parametrize workflow dispatch (manually trigger) to add models here. I will try this approach

scripts/pr_results_comment.py
Samoed: I thought if reference models won't have results then error will be risen, but I looked in code, and it seems that they will filter out, but I will try to test this case

scripts/pr_results_comment.py
Samoed: >   --files file1.json file2.json # if not specified, if will just get the diff from main (code in my script)

In that case, I don't think we need `--files` anymore. I thought it would be better to find diff with main from CI. I will move it then to main script

Samoed:
I've added support for max score and comparison with different models through `workflow_dispatch`, but I can't test it yet because it hasn't been merged

Samoed:
@KennethEnevoldsen Can you review this PR? Actual work of the script in this comment https://github.com/embeddings-benchmark/results/pull/220#issuecomment-2964037725

.github/workflows/model-results-comparison.yaml
```diff
@@ -0,0 +1,74 @@
...
+      with:
+        fetch-depth: 0
+
+    - name: Fetch origin main
+      run: git fetch origin main
+
+    - name: Set up Python
+      uses: actions/setup-python@v5
+      with:
+        python-version: '3.9'
```
KennethEnevoldsen: Let us use a newer version, as it will soon be deprecated

KennethEnevoldsen (approved):
Looks good! Feel free to merge

>To use it like a bot, we need to host it somewhere. I can host it, but I'm not sure if that's the best option. We could make it trigger on each comment and check if there’s a command to run, but I’m not sure if that’s a good idea.
>
> Another solution could be to parametrize workflow dispatch (manually trigger) to add models here. I will try this approach

I think for now it is perfectly fine to just have it as a default comment, and then we can run any additional stuff ourselves.

.github/workflows/model-results-comparison.yaml
Samoed: Bumped to `3.10`

PR #221, opened 2025-06-16
add Results of Hakims and Tooka-SBERTV2s

mehran-sarmadi:
In this PR, we add results for five models: three from the Hakim family (hakim, hakim-small, and hakim-unsup), and two from the TookaSBERTV2 family (TookaSBERT-V2-Small and TookaSBERT-V2-Large).

### Checklist

- [] My model has a reference implementation in [`mteb/models/`](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models) this can be as an API. Instruction on how to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) 
  - [x] No, but there is an existing PR https://github.com/embeddings-benchmark/mteb/pull/2826


mehran-sarmadi:
Hi @Samoed,
Since the issue mentioned here seems to be resolved (https://github.com/embeddings-benchmark/results/pull/222), would rerunning the compare-results check help, or should I make changes on my side?

Samoed:
I think you need to update your PR firstly, because restarting action would restart only old version. But for now I think you need to finish integration of your model

KennethEnevoldsen:
@Samoed, it seems like the CI fails here. Any idea why?

Samoed:
I think this action before all modifications and to update it main should be merged to this branch

KennethEnevoldsen:
@mehran-sarmadi many of these scores are waay above the baseline e5? Do you have any ideas why this might be? Especially for the `SynPerChatbot*` tasks

mehran-sarmadi:
> @mehran-sarmadi many of these scores are waay above the baseline e5? Do you have any ideas why this might be? Especially for the `SynPerChatbot*` tasks

Hi @KennethEnevoldsen 

The main reason for this is that the Hakim and Hakim-small models were fine-tuned on the **training** portions of the SynPerChatbot datasets during Stage 3 (supervised fine-tuning). As a result, their "zero-shot" scores on the leaderboard are much lower than those of models like E5 — for example, Hakim and Hakim-small score around 33%, whereas E5 scores approximately 91%. In contrast, the Hakim-unsupervised model did not undergo Stage 3 fine-tuning, and its zero-shot score is also around 90%, so its results are much closer to the E5 baseline — only slightly better, likely due to its exposure to a large volume of high-quality Persian data.

KennethEnevoldsen:
Thanks for the clarification @mehran-sarmadi, we will have to add this to the model card. You can add it like so:

```
# in model meta
...
training_datasets={"SynPerChatbotRAGTopicsRetrieval": ["train"], 
...} # add all datasets for which you have trained on (even if the split is the training split)
```

mehran-sarmadi:
> Thanks for the clarification @mehran-sarmadi, we will have to add this to the model card. You can add it like so:
> 
> ```
> # in model meta
> ...
> training_datasets={"SynPerChatbotRAGTopicsRetrieval": ["train"], 
> ...} # add all datasets for which you have trained on (even if the split is the training split)
> ```
Hi @KennethEnevoldsen 
Already done! They are included in the model meta.

KennethEnevoldsen:
Ah, sorry I overlooked that - then everything is good here!

PR #254, opened 2025-08-14
Add: PwC-Embedding_expr Results

ElPlaguister:
Submit results

### Checklist

- [x] My model has a reference implementation in [`mteb/models/`](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models) this can be as an API. Instruction on how to add a model can be found [PwC-Embedding_expr](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) 
  - [x] No, but there is an existing PR https://github.com/embeddings-benchmark/mteb/pull/3028


KennethEnevoldsen:
Nothing here looks especially problematic given the training data

PR #260, opened 2025-08-27
Add conversion of RTEB results for MTEB display

q275343119:
This PR converts the results from [RTEB](https://github.com/embedding-benchmark/rteb)
 into a compatible format so that they can be displayed within MTEB.

### Checklist


q275343119:
> This PR converts the results from [RTEB](https://github.com/embedding-benchmark/rteb) into a compatible format so that they can be displayed within MTEB.
> 
> ### Checklist
> * [x]  My model has a model sheet, report or similar
> * [x]  My model has a reference implementation in [`mteb/models/`](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models) this can be as an API. Instruction on how to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md)
>   
>   * [ ]  No, but there is an existing PR ___
> * [x]  The results submitted is obtained using the reference implementation
> * [x]  My model is available, either as a publicly accessible API or publicly on e.g., Huggingface
> * [x]  I _solemnly swear_ that for all results submitted I have not on the evaluation dataset including training splits. If I have I have disclosed it clearly.

Hi @KennethEnevoldsen ,could you please take a look at this PR when you  have time?

Samoed:
I don't think that this is good idea to convert your results to the mteb format, because they have a bit different metrics and different scores

q275343119:
> I don't think that this is good idea to convert your results to the mteb format, because they have a bit different metrics and different scores

I think we might need to consider this in two scenarios:
The first is for results that are not available in MTEB, and the second is for results that exist in both MTEB and this evaluation.

For the first scenario, would it be acceptable to submit them directly?
For the second scenario, what would you suggest?

Samoed:
> For the first scenario, would it be acceptable to submit them directly?

If they're not available then it's better to run them with mteb and submit them

> second is for results that exist in both MTEB and this evaluation.

Use mteb results

results/BAAI__bge-base-en-v1.5/a5beb1e3e68b9ab74eb54cfd186867f64f240e1a/AILACasedocs.json
KennethEnevoldsen, line 1: Conversion seems incorrect? mteb version is lower and dataset revision is not the full revision.
Eval time and CO2 consumption are also not specified.

These seem to be general (the main score also doesn't match?)

KennethEnevoldsen (changes requested):

results/BAAI__bge-base-en-v1.5/a5beb1e3e68b9ab74eb54cfd186867f64f240e1a/AILACasedocs.json
q275343119: Yes,you're absolutely right.we used the RTEB results directly instead of running the evaluation through the standard MTEB library.

We will re-run the evaluations properly using MTEB and submit a new PR with the corrected results shortly

KennethEnevoldsen: Alright - I will just close this PR then

PR #267, opened 2025-09-02
Adding more RTEB results

fzoll:
### Checklist

- [ ] I *solemnly swear* that for all results submitted I have not on the evaluation dataset including training splits. If I have I have disclosed it clearly.


fzoll:
Well, actually these are not my models, i just run MTEB and submitting the results.

PR #278, opened 2025-09-24
New results for ai-sage/Giga-Embeddings-instruct

ekolodin:
### Checklist


Samoed:
Hm, scores in table are lower than reported. I think my script for table generation averages scores between revisions. I will try to fix it later

Samoed:
For context, I contacted the authors but couldn’t reproduce their results. My results are actually higher, and they’re currently investigating this.

Samoed:
The new results match what I’m getting now.

PR #280, opened 2025-09-25
Fix table generation for models with multiple revisions

Samoed:
Close https://github.com/embeddings-benchmark/results/issues/279


scripts/create_pr_results_comment.py
```diff
@@ -107,47 +111,75 @@ def create_comparison_table(
...
+    # max_dataframe = task_results_df.groupby(task_col_name).max()
+    # high_model_performance_tasks = []
+    #
+    # model_select_colum = model if model in df.columns else f"{model}__{new_model_revision}"
+    # if not max_dataframe.empty:
+    #     for task_name, row in max_dataframe.iterrows():
+    #         df.loc[df[task_col_name] == task_name, max_col_name] = row["score"]
+    #         model_score = df.loc[df[task_col_name] == task_name, model_select_colum].values[0]
+    #         if model_score > row["score"]:
+    #             high_model_performance_tasks.append(task_name)
```
KennethEnevoldsen: del?

scripts/create_pr_results_comment.py
```diff
@@ -107,47 +111,75 @@ def create_comparison_table(
     max_col_name = "Max result"
     task_col_name = "task_name"
     results = mteb.load_results(models=models, tasks=tasks, download_latest=False)
-
-    results = results.join_revisions()
-    df = results.to_dataframe()
```
KennethEnevoldsen, line 112: This should not take the mean across (but it will prefer the revision in mteb)

KennethEnevoldsen: (but I think the change is nice)

KennethEnevoldsen (approved):

scripts/create_pr_results_comment.py
Samoed: I commented this code to fast create results table locally. I will roll back this

scripts/create_pr_results_comment.py
Samoed: It seems that during `to_dataframe` scores between revision are averaged 
https://github.com/embeddings-benchmark/mteb/blob/08bba494f03ab0933aa1fcdab59b848002069d30/mteb/load_results/benchmark_results.py#L50-L52

PR #281, opened 2025-09-27
Add human results

Samoed:
### Checklist


isaac-chung (approved):
For human results, we might need to specify the "aggregate" ones to be displayed.

Samoed:
> For human results, we might need to specify the "aggregate" ones to be displayed.

Yes, we would create model meta for this and we will specify split. But I think we need a better naming for this. Maybe with date of the results creation, e. g. `2025_09_27`. To have more control on them in future

isaac-chung:
Good call. Let's do that

KennethEnevoldsen (approved):
Looks good - not much to add besides what is already here

AdnanElAssadi56:
Added a script to fetch human annotation results from Argilla and save them in MTEB-compatible format.
We will need to run this from repo root, and it does the following:

- Fetches all annotations from Argilla API (even if datasets are only partially done so we need to be careful with this)
- Creates timestamped aggregate results in results/Human/{YYYY_MM_DD}/
- Updates individual annotator directories in results/Human/annotator_{id}/

Samoed:
@AdnanElAssadi56 Can you move out your script to separate PR? I think it would be easier to review it

KennethEnevoldsen:
Yeah let is remove the script here, and then I think we can merge this in?

isaac-chung (approved):
Script removed.

PR #286, opened 2025-10-03
Add static models to RTEB

KennethEnevoldsen:
Added static model evaluations to RTEB!


KennethEnevoldsen (commented):
I don't see any problematic diffs - all results are either new, exactly the same or below or with a diff less than 0.01 (typically notably lower)

I sadly didn't run it with CO2 emissions on so this values is set to null when overwriting results, but I think it is better to have results using a newer version than CO2 emissions (but will make sure to add it for future runs)

PR #287, opened 2025-10-04
add Jina-v4 to RTEB

KennethEnevoldsen:
Added results for Jina's v4 to RTEB. Some of the results doesn't match previous results.

related to: https://github.com/embeddings-benchmark/mteb/issues/3238

PR #288, opened 2025-10-04
Adding Jina V4 results

fzoll:
Add more results related to RTEB

### Checklist


KennethEnevoldsen:
@fzoll, can you remove the Jina-v4 results? There is a PR on it here #287, which looks into a few performance discrepancies. Otherwise I think we can merge this

makram93:
Hi @KennethEnevoldsen, thanks for merging the other PR #287 👍 
I noticed that only English tasks were included there. Could we keep the missing evaluation results, including the multilingual ones, done in this PR for jina-v4?

KennethEnevoldsen:
@makram93 totally - let us just do that

KennethEnevoldsen (approved):

PR #289, opened 2025-10-04
score bm25 on RTEB

Samoed:
### Checklist

- [X] My model has a model sheet, report or similar
- [X] My model has a reference implementation in [`mteb/models/`](https://github.com/embeddings-benchmark/mteb/tree/main/mteb/models) this can be as an API. Instruction on how to add a model can be found [here](https://github.com/embeddings-benchmark/mteb/blob/main/docs/reproducible_workflow.md) 
- [X] The results submitted is obtained using the reference implementation
- [X] My model is available, either as a publicly accessible API or publicly on e.g., Huggingface


Samoed:
Wow, bm25 have the highest performance on 3 tasks

KennethEnevoldsen:
loving a strong baseline ;)

PR #290, opened 2025-10-06
fix all r-paris & r-oxford results for MIEB leaderborad

gowitheflow-1998:
fixing issues revealed by https://github.com/embeddings-benchmark/mteb/issues/3256

1. Delete all r-Oxford and r-Pairs `Retrieval` results (old wrong implementation).
2. Rename all r-Oxford and r-Pairs `MultiChoice` results into `Retrieval` (correct implementation that matches the paper results).

deleted 270 and renamed 276 in total (the 6 should be what's missing in https://github.com/embeddings-benchmark/mteb/issues/3256).


isaac-chung (approved):
Thanks!

PR #292, opened 2025-10-07
update youtu embedding model metadata

spring-quan:
### Checklist


Samoed:
@spring-quan You need to rename folder too

results/tencent__Youtu-Embedding/1/model_meta.json
Samoed, line 1: Now you need either rename model here or don't rename this revision

Samoed (approved):

PR #302, opened 2025-10-17
MMTEB results for llama-embed-nemotron-8b

ybabakhin:
Adds MMTEB results for [llama-embed-nemotron-8b](https://huggingface.co/nvidia/llama-embed-nemotron-8b) model

### Checklist

  - [x] No, but there is an existing PR: https://github.com/embeddings-benchmark/mteb/pull/3407


Samoed:
Tasks with improvements over previous max

1. `MIRACLRetrievalHardNegatives` +3
2. `ScalaClassification` +7
3. `SIB200ClusteringS2S` +3
4. `IndonesianIdClickbaitClassification` +7

KennethEnevoldsen:
`MIRACLRetrievalHardNegatives` is in the training set. Scala can be a bit sporadic so wouldn't worry too much about that.

`IndonesianIdClickbaitClassification` might be a case of a better prompt (the current ones seem quite bad; we should probably update that one). I see that SIB200 also have custom prompt.

@ybabakhin can I ask about the approach that you used for selecting prompts (I see that not all tasks have them).

Samoed:
> the current ones seem quite bad; we should probably update that one

We can't update prompts, because results wouldn't be reproducible

ybabakhin:
@KennethEnevoldsen approach was the following:
* Take prompts from `Qwen3-Embedding` models: https://github.com/QwenLM/Qwen3-Embedding/blob/main/evaluation/task_prompts.json
    * For example, it already had `"IndonesianIdClickbaitClassification": "Given an Indonesian news headlines, classify its into clickbait or non-clickbait",`
* Edit a general prompt for Retrieval datasets: from `Retrieval the relevant passage for the given query` -> `Given a question, retrieve passages that answer the question`
* Remove datasets which already have own prompt in `mteb`
* Remove datasets for STS, BitextMining and PairClassification problem types, as they all have general prompts: https://github.com/embeddings-benchmark/mteb/pull/3407/files#diff-cbe979b0366cc5f6a7bbc0f0cf20f494873d162415e68d2b80e0f2bd9a28ce98R437

KennethEnevoldsen:
Thanks for the clarification @ybabakhin. I will merge this in as it doesn't seem like there are any notable changes in the model implementation.

Btw @ybabakhin did you consider running this on RTEB as well? (we can help run the private set if you run the public)

> We can't update prompts, because results wouldn't be reproducible

We can introduce a new version of the tasks and rerun existing set of models

ybabakhin:
@KennethEnevoldsen Thanks for merging!

> Btw @ybabakhin did you consider running this on RTEB as well?

Unfortunately, RTEB was released when we already had finalized our data mix. We don't have any Code data in there, while RTEB is pretty code-heavy. So, we're not targeting it with this model.

But we like a Public-Private Leaderboard setup, and we can re-visit RTEB in the future

PR #306, opened 2025-10-24
fix test for v2

Samoed:


.github/workflows/test.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest] #, macos-latest, windows-latest]
-        python-version: ["3.9"]
+        python-version: ["3.10"]
```
Samoed, line 19: We need to update required CI to use python3.10 instead 3.9

Samoed: Interesting that mteb `v2` was installed on `3.9`

KennethEnevoldsen:
fixed

PR #307, opened 2025-10-24
Add results for MTEB-NL

nikolay-banar:
### Checklist


Samoed (approved):

Samoed:
I can't merge for now, because test are failing, but we have fix in https://github.com/embeddings-benchmark/results/pull/306

PR #309, opened 2025-10-29
Refreshing the voyage-multimodal-3 Vidore results

fzoll:
Refreshing the `voyage-multimodal-3` `Vidore` results

### Checklist


KennethEnevoldsen (commented):
Hmm why does the scores change here? 

Has the model implementation changed without a revision bump?

fzoll:
@KennethEnevoldsen Let's pick `VidoreTatdqaRetrieval` as an example. In 2025.02.04 (feat: Merge MIEB into main 🎉  (#1944)) merge, the task started to use the actual revision. But previously (since 2024.10.15) a different path was used. The difference is the format, the new version (if i understand correctly) has deduplicated documents: the number of documents dropped from 1663 to 277, resulting that models performing better on the new version. The previous `voyage-multimodal-3` results were added on 2025.02.23, but the dataset revision number in the JSON is the original.

  What Changed

  Original (Oct 15, 2024 - Dec 25, 2024):
  - Dataset path: "vidore/tatdqa_test"
  - Revision: "9c3a626c16c811f15514689c3e7e95a4f2b9b8c3"
  - Number of documents: 1,663
  - Number of queries: 1,663

  Current (Dec 25, 2024 - present):
  - Dataset path: "vidore/tatdqa_test_beir"
  - Revision: "5feb5630fdff4d8d189ffedb2dba56862fdd45c0"
  - Number of documents: 277
  - Number of queries: 1,663 (unchanged)

Do i see something wrong here?
I'm planning to run other models against this new revision as well, checking which models has the scores from the previous dataset version - that's why i changed this PR to draft.

Samoed:
I think we need to check other results with old dataset revision. CC @isaac-chung

fzoll:
@Samoed I just checked, but i did not find any result file with the old revision number (9c3a626c16c811f15514689c3e7e95a4f2b9b8c3)

Samoed:
I can't either. It seems that this is only model that wasn't reevaluated

KennethEnevoldsen:
I will just ping @isaac-chung again here, but it looks like we can merge this regardless

KennethEnevoldsen (approved):

isaac-chung:
Hey all, 
Apologies for being late to the party. 

Thanks for updating the scores. It is very likely that when we [switched the vidore dataset paths last Xmas](https://github.com/embeddings-benchmark/mteb/commit/ff74380542e3942b7e60e782636dc2aa7a4c0d3e), the previous scores were not updated. The dataset name remained the same and that likely was the cause for the missing rerun. 

Voyage multimodal 3 is already 1st in the original MIEB and it doesn't seem like the change in scores affects ranking. This PR looks good to me.

PR #318, opened 2025-11-14
add Jasper-Token-Compression-600M results on mteb(eng, v2)

DunZhang:
### Checklist

  - [x] No, but there is an existing PR  https://github.com/embeddings-benchmark/mteb/pull/3557


Samoed (approved):
I think overall looks good, but on sts and classification tasks scores are high

DunZhang:
> I think overall looks good, but on sts and classification tasks scores are high

@Samoed 
Hello, this is a characteristic of vector distillation. We recently found that in multi-teacher distillation, for classify clustering tasks(Symmetrical), the student model can almost always easily approach the performance of the teacher models, and even surpass individual teachers. However, on retrieval tasks, the gap is much larger and it’s very hard to fit.

In our current open-source model, we even combined contrastive learning with distillation training to further strengthen retrieval performance, but it only brought a slight improvement.

In short, this is a normal phenomenon, because the two teacher models (qwen3-8b and qzhou) themselves perform very well on both classification and sts.

KennethEnevoldsen:
Thanks for the clean PRs @DunZhang - models results should be up tomorrow with the automatic update of the leaderboard

PR #364, opened 2025-12-13
Remove duplicate results

ayush1298:
Related to https://github.com/embeddings-benchmark/mteb/issues/3383

### Checklist

- [ ] The results submitted are obtained using the reference implementation


ayush1298:
@Samoed Can you check the script once, If its as expected then I will run it

clean_duplicate_results.py
Samoed, line 1: Can you create file with models/revisions that will be deleted?

ayush1298: 1 doubt, when only external and no_revision_available is present, then should I keep external and delete no_revision_available?

Samoed: It depends. Some models could have most results in external

ayush1298:
@Samoed Can you check the csv?

clean_duplicate_results.py
ayush1298: There is just 1 models, with only 2 revisions, external and no_revision_available.  jhu-clsp/FollowIR-7B ( Both revisions have 3 task results only), So, I will take external and remove other one.

Samoed (commented):
For `Alibaba-NLP/gte-multilingual-base`, `NovaSearch/jasper_en_vision_language_v1`, `intfloat/multilingual-e5-large`, `openai/text-embedding-3-small` and some more   seems to be a bit strange because same revision shoulve retained and deleted. Human results shodn't be deleted

deleted_revisions.csv
```diff
@@ -0,0 +1,179 @@
+model,revision_retained,deleted_revisions,num_deleted_revisions
+Alibaba-NLP/gte-Qwen1.5-7B-instruct,07d27e5226328010336563bc1b564a5e3436a298,"external, no_revision_available",2
```
KennethEnevoldsen: You can't just delete the entire revision. It is only if the task exists in the reference revision that you can delete the results for only that one task.

KennethEnevoldsen (commented):
@ayush1298 please spend longer reviewing these PRs before you ask for input. You haven't even resolved the github copilot comments.

While we are more than happy to give feedback, do read through it first and think what would be a question that a reviewer might have and supply that information.

I do really appreciate the effort that you put in, but I would really appreciate if the initial quality was higher as otherwise we spend more time reviewing that it would be to do this ourselves.

I think the best way forward is here to add logging to the script. Run it while the delete part is commented out and then add the logfile to the PR. Log messages should be something like

```
===Deduplicating model (Cohere/Cohere-embed-multilingual-v3) ===
reference revision "c6c1b92f4a3e1b92b326ad29dd3c8433457df8dd"

Found duplicates for Task
    deleting c167e9a8144b397622ce47b85d9edcdeecef3d3f/Task.json as c6c1b92f4a3e1b92b326ad29dd3c8433457df8dd/Task.json already exists
    ...
```
It should be clear what is deleted and why.

deleted_revisions.csv
ayush1298: I am little bit confused here, so, are you saying that if for 1 task, results are in both `reference` revision, and in some other revision(e.g` external`), then we can just delete results of that particular task from `external`(will not delete the whole `external` revision). But, then at the end, isn't it also possible that there will be some tasks whose results will not be in `reference` revision, but they will be in `external `revision, and we still end up having 2 revisions. 

In the leaderboard, we use results based on revision, right? Or we will go to the model, and then go to the task, and then take the results of any revision at random?

Also, I will add more log file and anything I have assumed so that it will be easier for review

ayush1298:
> @ayush1298 please spend longer reviewing these PRs before you ask for input. You haven't even resolved the github copilot comments.
> 
> While we are more than happy to give feedback, do read through it first and think what would be a question that a reviewer might have and supply that information.
> 
> I do really appreciate the effort that you put in, but I would really appreciate if the initial quality was higher as otherwise we spend more time reviewing that it would be to do this ourselves.

Sorry, @KennethEnevoldsen . I would try to be more careful from next time.

deleted_revisions.csv
KennethEnevoldsen: > But, then at the end, isn't it also possible that there will be some tasks whose results will not be in reference revision, but they will be in external revision, and we still end up having 2 revisions.

Yes that is the case - we would rather have two revision rather than delete a task from a model

> In the leaderboard, we use results based on revision, right? Or we will go to the model, and then go to the task, and then take the results of any revision at random?

In the leaderboard we prefer one revision if there is multiple task results, but if there is no result for Task A in the reference revision then we use another one

> Also, I will add more log file and anything I have assumed so that it will be easier for review

Thanks - again I do really appreciate these PRs (do resolve comments from the copilot and us who you believe are resolved)

ayush1298:
I have added log file and updated script.
1 thing that I consider is whenever some task is not present in reference revision, and its present in some other revisions, then I am keeping that task from the revision based on priority:  `any valid revision > external > no_revision_available`. So, if some task is present in both `external `and `no_revision_available,` then I will keep the file from `external` and delete from `no_revision_available`.

I have also excluded Human folder from these deletion. 
I have also checked results for some models like `openai/text-embedding-3-small`,  `Alibaba-NLP/gte-Qwen1.5-7B-instruct` and it seems fine to me.

ayush1298:
@Samoed Can you now check? Also, let me know if there is confusion at anything. I have summarized things above.

Samoed (commented):
I think script will work correctly. Checked logs

ayush1298:
> I think script will work correctly. Checked logs

@Samoed I have checked the logs for 2-3 models, and it seems correct to me.  Should I run the script and perform deletion?

Samoed:
Let's wait for @KennethEnevoldsen review

KennethEnevoldsen:
Reviewed the log file it seems fine to me - thanks for taking the time!

ayush1298:
@Samoed I have 1 more doubt here, currently In a script, I am deleteing folder if It contains just model_meta.json file, or even if folder contains other duplicate files which gets deleted and it just ended with model_meta.json file:  like https://github.com/embeddings-benchmark/results/tree/main/results/google__flan-t5-base/7bcac572ce56db69c1ea7c8af255c5d7c9672fc2

Is this correct or not? Or we have to keep these folders?

Samoed:
I think you can delete them

ayush1298:
> I think you can delete them

Do we use model_meta.json anywhere in leaderboard. Asking these because, https://github.com/embeddings-benchmark/results/tree/main/results/jhu-clsp__FollowIR-7B, for this, script is keeping no_revision_available folder and removing external folder. And model_meta.json is present only in external folder and not in no_revision_available folder

Samoed:
It will be used if `model_meta.json` is required in cache https://github.com/embeddings-benchmark/mteb/blob/6298c86816d999bfd163324a24d737e31be78c33/mteb/cache.py#L326, but for leaderboard building we're disable this https://github.com/embeddings-benchmark/mteb/blob/6298c86816d999bfd163324a24d737e31be78c33/mteb/leaderboard/app.py#L47. But I can't find `FollowIR-7b` on leaderboard http://mteb-leaderboard.hf.space/?benchmark_name=FollowIR

ayush1298:
> It will be used if `model_meta.json` is required in cache https://github.com/embeddings-benchmark/mteb/blob/6298c86816d999bfd163324a24d737e31be78c33/mteb/cache.py#L326, but for leaderboard building we're disable this https://github.com/embeddings-benchmark/mteb/blob/6298c86816d999bfd163324a24d737e31be78c33/mteb/leaderboard/app.py#L47. But I can't find `FollowIR-7b` on leaderboard http://mteb-leaderboard.hf.space/?benchmark_name=FollowIR

same, maybe some other issue with it. 

Also, for model_meta.json, so, should I keep those folders even if they just contain single file which is model_meta.json? And also for: 

> https://github.com/embeddings-benchmark/results/tree/main/results/jhu-clsp__FollowIR-7B, for this, script is keeping no_revision_available folder and removing external folder. And model_meta.json is present only in external folder and not in no_revision_available folder

though script maintains priority order, but when only external and no_revision_available is present, it is I think doing it randomly.

Samoed:
> should I keep those folders even if they just contain single file which is model_meta.json? And

No, you can delete them

> I think doing it randomly.

I think it's better to keep no_revision_available

ayush1298:
> > should I keep those folders even if they just contain single file which is model_meta.json? And
> 
> No, you can delete them
> 
okay. 

> > I think doing it randomly.
> 
> I think it's better to keep no_revision_available
>

oaky, so, even if there is no model_meta.json in no_revision_available, we are good, right?

Samoed:
> oaky, so, even if there is no model_meta.json in no_revision_available, we are good, right?

yes

ayush1298:
Removed results and also removing script and log file.

Samoed (approved):

ayush1298:
@KennethEnevoldsen Could you review this one whenever you get time? I think its done

KennethEnevoldsen:
Great PR @ayush1298, lovely to reduce the number of models! It would be great to add a test on this as well (to prevent it going forward), but I will leave that for another PR

PR #374, opened 2025-12-23
add RTEB & MTEB results of Octen-Embedding-8B

bflhc:
### Checklist

  - [x] No, but there is an existing PR https://github.com/embeddings-benchmark/mteb/pull/3789


bflhc:
Octen-Embedding-8B is optimized for retrieval tasks. It is trained on Qwen3-Embedding-8B using a large amount of real-world industry search data, combined with high-quality synthetic data.

From the results, we can see significant improvements over the base model on both RTEB and MTEB reranking/retrieval tasks. We believe this model could bring substantial value to the open-source community.

We would also appreciate it if you could help run the RTEB private evaluation, so that we can more comprehensively assess the model’s performance.

Regarding configuration, a `batch_size` of 4 or 2 should work well. Setting `corpus_chunk_size` to 5000 can also be helpful for datasets with a very large number of queries.

Samoed (approved):
Run Aila tasks and can reproduce results

PR #377, opened 2025-12-24
Fix GitHub workflow: prevent massive commits in cached-data branch

isaac-chung:
Fixes the issue where the 'generate_cached_results.yml' workflow was committing 84,642+ files instead of just the cache file, causing push failures due to GitHub limits.

Changes:
- Properly handle orphan branch creation by cleaning staging area
- Add safety check to prevent commits with >10 files
- Improve branch existence detection and handling
- Better file management for README.md preservation
- More robust error handling throughout the process

This should fix the "Process completed with exit code 1" error in GitHub Actions runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Samoed:
That's strange. Why so much files where committed if you're using only `git add __cached_results.json.gz`

isaac-chung:
The original issue was that git checkout --orphan cached-data copies all files from the current branch to the staging area. The script wasn't properly cleaning this up, so when it added the cache file, it was committing everything (84,642 files instead of just 1).

Samoed (approved):

PR #390, opened 2026-01-06
update bflhc/Octen-Embedding-8B RTEB results

bflhc:
### Checklist


bflhc:
The MIRACLRetrievalHardNegatives evaluation is still running; results are expected to be updated in about 3 hours.
This is a new revision of bflhc/Octen-Embedding-8B. The corresponding PR is at https://github.com/embeddings-benchmark/mteb/pull/3869. Could you please merge the PR and also trigger the private eval afterward? Thanks! @Samoed

bflhc:
I have updated all the public results. Could you please merge this PR?
If possible, would also be helpful to share an update on the progress of the private evaluation. @Samoed

Samoed (approved):

Samoed:
Evaluation of 8b model takes too much time and resources. I don't think you would get huge bump

bflhc:
> Evaluation of 8b model takes too much time and resources. I don't think you would get huge bump

Thanks a lot for the honest feedback — I totally understand the concern about 8B private eval being expensive.
I just want to briefly clarify why I believe this particular run could still be meaningful for RTEB, even if the absolute bump turns out to be moderate.

  1. This is a controlled comparison on the same 8B backbone, not a scale-up experiment. The architecture and inference path are unchanged; the only difference is the training data and objectives. From a benchmark perspective, this makes the result easier to interpret.
  2. The model is specifically optimized for specific-domain and long-context retrieval, including legal / government-style data, which is exactly where several RTEB private tasks (e.g. legal and regulation-related datasets) differ most from public benchmarks. These effects may not fully surface on public MTEB/RTEB.
  3. On public benchmarks, we already observe consistent improvements over the base model on retrieval-heavy tasks, which suggests the changes are not trivial.

Even if the gains are modest, I think having a well-controlled 8B result could still be useful signal for the community, especially for people working on domain-adapted embeddings.

PR #399, opened 2026-01-11
Pass ModelMeta objects to filter by revision when generating cache

isaac-chung:
## Summary
- Pass `ModelMeta` objects instead of model name strings to `load_results()`
- This ensures filtering uses both model name AND revision
- Avoids loading `no_revision_available` folders when a proper revision exists

## Problem
The generated cache contained entries from `no_revision_available` folders with `dataset_revision: null`, causing 458 pydantic validation errors when loading with `BenchmarkResults.from_disk()`.

Fixes https://github.com/embeddings-benchmark/mteb/issues/3916

PR #400, opened 2026-01-13
Revert "Pass ModelMeta objects to filter by revision when generating cache"

Samoed:
Reverts embeddings-benchmark/results#399 from discussion https://github.com/embeddings-benchmark/mteb/issues/3918

### DIFF
--- a/scripts/generate_cached_results.py
+++ b/scripts/generate_cached_results.py
@@ -40,24 +40,18 @@ def generate_cached_results():
     # The remote repo should already be cloned from previous runs
     logger.info("Using existing remote results repository...")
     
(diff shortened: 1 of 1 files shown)


KennethEnevoldsen (approved):

KennethEnevoldsen (commented):
This fixed:
https://github.com/embeddings-benchmark/mteb/issues/3916

Does that re-open the issue?

Samoed:
I don't fully understand how this would fix this problem. I can load old cache without problems

KennethEnevoldsen:
otherwise we have the fallback as well

PR #413, opened 2026-02-01
add all MAEB results

### LINKED ISSUE #4021: Add MAEB results to results repo
Currently they are on their own repo.

isaac-chung:
Fixes https://github.com/embeddings-benchmark/mteb/issues/4021

- Copy over results from https://github.com/embeddings-benchmark/maeb-results/tree/main
- I'd keep the `maeb-results` repo in archive in just case.


KennethEnevoldsen:
right now tests are failing due to the lack of implementation in `mteb`, I would suggest waiting until we have the MAEB merge complete, but simply keep working with results in this branch (such that this branch is now the reference for MAEB). WDYT?

@AdnanElAssadi56, @Samoed, @gowitheflow-1998  are there any of the results that are problematic to merge here?

Samoed:
> there any of the results that are problematic to merge here?

What do you mean problematic?

KennethEnevoldsen:
I know that we have changed some tasks and I want to make sure that we don't have any results that are not invalidated by this

Samoed:
I think would be great to reevaluate results on MAEB, because this is hard to track everything

KennethEnevoldsen:
Yeah, we might do it with just a small models (to find problematic tasks) and then we can rerun those on all models

isaac-chung:
> due to the lack of implementation in mteb

Let's at least wait for this first then.

isaac-chung:
At least the tests and the file size check are passing. Will take a look at comparison separately.

PR #423, opened 2026-02-25
[ADD] additional resuls from vie-vn models from VN-MTEB benchmark

BaoLocPham:
### Checklist


Samoed (changes requested):
Hi! We don't have in mteb `LegalNewsVNRetrieval` can you add it to repo or remove from results? Also, we don't have implementations for some models. And you added results in multiple revisions for some models (`gemma-embeddings-300m`, e5 models)

PR #431, opened 2026-03-06
Add ColBERT-Zero and GTE-ModernColBERT-v1 results

NohTow:
I am continuing to add the results of our ColBERT models, this time with the Zero series and GTE-ModernColBERT-v1
### Checklist

https://huggingface.co/collections/lightonai/colbert-zero
https://huggingface.co/lightonai/GTE-ModernColBERT-v1
GTE-ModernColBERT-v1 is already there, I'm making a PR for the zero serie just after this PR
The only specific parameters to change from defaut loading are lengths. For longembed, we need to set doclength to 32k to match the bench.
For BEIR, I am still using different query lengths per dataset to match what was done by old models, but honestly now that we do not have query expansion anymore, it can be set to whatever fits all the queries.
https://huggingface.co/collections/lightonai/colbert-zero
https://huggingface.co/lightonai/GTE-ModernColBERT-v1
ColBERT-Zero models are trained on exclusively public data. GTE-ModernColBERT-v1 is only fine-tuned on MS MARCO and is based on the GTE model that have stated their data in the paper/is already merged here.


NohTow:
Also, I compared my results reported back then to those ones, to check that they match
There are obviously small deltas but it seems very fair, at least for BEIR.
LongEmbed is a bit different, but I believe it might be due to how draw are handled in the MTEB evaluator, which seems very helpful for ColBERT models on some datasets when using quantified indexes. Running the tests are pretty quicky.
image

Samoed:
Great! Can you add `ModelMeta` of `lightonai/ColBERT-Zero` models?

NohTow:
Done!

KennethEnevoldsen (approved):
assuming tests pass then this looks good!

PR #443, opened 2026-03-13
Rerun qwen results

KennethEnevoldsen:


KennethEnevoldsen (commented):
This is main to resolve the evaluation time. This was run on the same hardware but with flash attention enabled

PR #444, opened 2026-03-13
Create Results PR Comment for Results diff

ayush1298:
Related to issue: https://github.com/embeddings-benchmark/mteb/issues/3595

This PR will basically check if results of a model gets updated in PR. If it is, then it will create a comment with diff of old and new results. 

I am not sure whether we want just a diff comment or we want to convert it to test as well where we can check diff for all changed results is under certain threshold. Maybe we can keep margin of 2-3%.


ayush1298:
I have updated the results of` intfloat/multilingual-e5-large` just for testing purposes. Will revert them before merge.

ayush1298:
@Samoed Here I have updated the YAML file, but its target is main branch, so it's running old command only and giving error and no table is created. So, what should I do?

Samoed:
You can try to run in your fork or create different action to test this. Or you can submit output of your script here

ayush1298:
# Updated Results: Old vs New Comparison

**Comparing against:** `origin/main`

**Total metric changes:** 7

### intfloat/multilingual-e5-large

| model_name                     | revision                                 | task_name               | split   | hf_subset   | metric      |   old_value |   new_value |   delta | pct_change   |
|:-------------------------------|:-----------------------------------------|:------------------------|:--------|:------------|:------------|------------:|------------:|--------:|:-------------|
| intfloat/multilingual-e5-large | 0dc5580a448e4284468b8909bae50fa925907bc5 | ESCIReranking           | test    | jp          | main_score  |    0.83202  |    0.87202  |    0.04 | +4.81%       |
| intfloat/multilingual-e5-large | 0dc5580a448e4284468b8909bae50fa925907bc5 | ESCIReranking           | test    | jp          | map_at_1    |    0.13655  |    0.23655  |    0.1  | +73.23%      |
| intfloat/multilingual-e5-large | 0dc5580a448e4284468b8909bae50fa925907bc5 | ESCIReranking           | test    | jp          | ndcg_at_1   |    0.81568  |    0.83568  |    0.02 | +2.45%       |
| intfloat/multilingual-e5-large | 0dc5580a448e4284468b8909bae50fa925907bc5 | ESCIReranking           | test    | jp          | recall_at_1 |    0.13655  |    0.23655  |    0.1  | +73.23%      |
| intfloat/multilingual-e5-large | 4dc6d853a804b9c8886ede6dda8a073b7dc08a81 | AfriSentiClassification | test    | amh         | f1          |    0.386177 |    0.446177 |    0.06 | +15.54%      |
| intfloat/multilingual-e5-large | 4dc6d853a804b9c8886ede6dda8a073b7dc08a81 | AfriSentiClassification | test    | amh         | main_score  |    0.432716 |    0.492716 |    0.06 | +13.87%      |
| intfloat/multilingual-e5-large | ab10c1a7f42e74530fe7ae5be82e6d4f11a719eb | AlloProfClusteringS2S   | test    | default     | main_score  |    0.322649 |    0.372649 |    0.05 | +15.50%      |

ayush1298:
Above one is generated `test-diff.md `. I am not adding comparison one as no new model is added.

Samoed:
I think you report only main score. I think better to change to format

| Task name | model name | model name | delta |
|-|-|-|-|
| Revision | 1 | 2 ||
| Task | 0.5 | 0.6 | +0.1 |

ayush1298:
| model_name      | task_name                 | old_revision            |   old_score |   new_revision |     new_score |   delta 
|:----------------|:-----------------------|:----------------|:--------|:------------|:------------|------------:|
| intfloat/multilingual-e5-large | ESCIReranking           |  revision_old |    0.83202  |   revision_new |  0.87202  |    0.04 | 

I think we need can show both old and new revision as these way even they have created some new file with different revision, then they can get diff. And anyway, they have to remove the duplicate task results, so from here they can also know which revision has same task results. 

Why should we don't show the other metrics and only main_score?
Also, do we have to convert these to some tests where we check if delta is under certain threshold?

KennethEnevoldsen:
> Why should we don't show the other metrics and only main_score?

Let us focus on main score for now

> Also, do we have to convert these to some tests where we check if delta is under certain threshold?

I am actually not sure we need the comment, but we def. need a test that fails.

ayush1298:
> > Why should we don't show the other metrics and only main_score?
> 
> Let us focus on main score for now
> 
> > Also, do we have to convert these to some tests where we check if delta is under certain threshold?
> 
> I am actually not sure we need the comment, but we def. need a test that fails.

I think comment is better for viewing the results diff exactly. In tests, so it would be based on some threshold right? maybe 2-3% change?

Samoed:
> I think we need can show both old and new revision as these way even they have created some new file with different revision

I meant like

| task name     | intfloat/multilingual-e5-large | intfloat/multilingual-e5-large | delta |
|:--------------|:-------------------------------|:-------------------------------|------:|
| Revisions     | old revision                   | new revision                   |       |
| Escireranking | 0.83202                        | 0.87202                        |  0.04 |
| Task 2 | 0.8                        | 0.6                        |  0.2 |

I don't thin we need to duplicate model name on each line, because this information just repeats

ayush1298:
> I don't thin we need to duplicate model name on each line, because this information just repeats

what about case when we have more than 1 model whose results changes?

Samoed:
Another table like in results PR now

ayush1298:
# Updated Results: Old vs New Comparison

**Comparing against:** `origin/main`

**Total tasks updated:** 5

| Model | Task | Old Revision | Old Score | New Revision | New Score | Δ Score | % Change |
|-------|------|--------------|-----------|--------------|-----------|---------|----------|
| intfloat/multilingual-e5-base | AfriSentiClassification | d13f1b27baf31030b7fd040960d60d909913633f | 0.433617 | d13f1b27baf31030b7fd040960d60d909913633f | 0.453617 | +0.020000 | +4.61% |
|  | JaCWIRRetrieval | 835193815a3936a24a0ee7dc9e3d48c1fbb19c55 | 0.842450 | 835193815a3936a24a0ee7dc9e3d48c1fbb19c55 | 0.862450 | +0.020000 | +2.37% |
| intfloat/multilingual-e5-large | AfriSentiClassification | 4dc6d853a804b9c8886ede6dda8a073b7dc08a81 | 0.432716 | 4dc6d853a804b9c8886ede6dda8a073b7dc08a81 | 0.452716 | +0.020000 | +4.62% |
|  | AlloProfClusteringS2S | ab10c1a7f42e74530fe7ae5be82e6d4f11a719eb | 0.322649 | ab10c1a7f42e74530fe7ae5be82e6d4f11a719eb | 0.352649 | +0.030000 | +9.30% |
|  | ESCIReranking | 0dc5580a448e4284468b8909bae50fa925907bc5 | 0.832020 | 0dc5580a448e4284468b8909bae50fa925907bc5 | 0.882020 | +0.050000 | +6.01% |

ayush1298:
@Samoed Will something like these works instead of multiple different table?

Samoed:
I think this can work, but in that case I think would be better to move revisions to the last columns to have better overview of scores

ayush1298:
Update table format and added tests. In tests, I am checking if % change is within threshold (kept 5% for now). If not, then test fails

Added table in below comment

ayush1298:
# Updated Results: Old vs New Comparison

**Comparing against:** `origin/main`

**Total tasks updated:** 5

| Model | Task Name | Old Score | New Score | Δ Score | % Change | Old Revision | New Revision |
|-------|-----------|-----------|-----------|---------|----------|--------------|--------------|
| intfloat/multilingual-e5-base | AfriSentiClassification | 0.433617 | 0.453617 | +0.020000 | +4.61% | d13f1b27baf31030b7fd040960d60d909913633f | d13f1b27baf31030b7fd040960d60d909913633f |
|  | JaCWIRRetrieval | 0.842450 | 0.862450 | +0.020000 | +2.37% | 835193815a3936a24a0ee7dc9e3d48c1fbb19c55 | 835193815a3936a24a0ee7dc9e3d48c1fbb19c55 |
| intfloat/multilingual-e5-large | AfriSentiClassification | 0.432716 | 0.452716 | +0.020000 | +4.62% | 4dc6d853a804b9c8886ede6dda8a073b7dc08a81 | 4dc6d853a804b9c8886ede6dda8a073b7dc08a81 |
|  | AlloProfClusteringS2S | 0.322649 | 0.352649 | +0.030000 | +9.30% | ab10c1a7f42e74530fe7ae5be82e6d4f11a719eb | ab10c1a7f42e74530fe7ae5be82e6d4f11a719eb |
|  | ESCIReranking | 0.832020 | 0.882020 | +0.050000 | +6.01% | 0dc5580a448e4284468b8909bae50fa925907bc5 | 0dc5580a448e4284468b8909bae50fa925907bc5 |

scripts/create_pr_results_comment.py
Samoed, line 1: I think better to move your script to another file

.github/workflows/model-results-comparison.yaml
```diff
@@ -87,3 +93,13 @@ jobs:
...
+
+    - name: Post diff PR comment
+      if: steps.pr_info.outputs.pr_number != ''
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      run: |
+        echo "Posting results diff comment..."
+        head -c 65300 model-diff.md > model-diff-truncated.md
+        echo -e "\n\n---\n**Note:** Content truncated due to GitHub API limits. See the full report in the workflow artifacts." >> model-diff-truncated.md
+        gh pr comment ${{ steps.pr_info.outputs.pr_number }} --body-file model-diff-truncated.md
```
Samoed: This would create problems in comment generation:
1. User submit results
2. Model results comment is created & diff
3. User update results
4. Comment with diff is now `results comment` (because of `--edit-last`) and new diff comment will be created

scripts/create_pr_results_comment.py
```diff
@@ -63,6 +69,171 @@ def get_diff_from_main() -> list[str]:
...
+
+    if diff_df.empty:
+        parts.append("No comparable updated result files found.")
+        return "\n".join(parts)
+
+    parts.append(f"**Total tasks updated:** {len(diff_df)}")
+    parts.append("")
+    
+    # Build markdown table with merged model_name rows, revisions on right
+    table_rows = [
```
Samoed: Why not simply `df.to_markdown`?

scripts/create_pr_results_comment.py
```diff
@@ -63,6 +69,171 @@ def get_diff_from_main() -> list[str]:
...
+        except (json.JSONDecodeError, IOError, KeyError):
+            continue
+
+        # Load old version from base ref
+        old_json = load_json_from_git_ref(relative_path, base_ref)
+        if old_json is None:
+            continue
+
+        # Load old revision from model_meta.json in base ref
+        old_model_meta = load_json_from_git_ref(str(model_meta_path.relative_to(repo_path)), base_ref)
```
Samoed: I think it's better to use `TaskResult` here

Samoed (commented):
We need to create this comment only if some results have changed, but now actions will be triggered every time

scripts/create_pr_results_comment.py
ayush1298: Just for formatting purposes. Like to skip having model_name in multiple results

scripts/create_pr_results_comment.py
ayush1298: Any purpose of this one, or just to keep code clean?

ayush1298:
> We need to create this comment only if some results have changed, but now actions will be triggered every time

Added check for this.

.github/workflows/model-results-comparison.yaml
ayush1298: We can add markers for both type of comment and then update exact comment. Will, that work?

scripts/create_pr_results_comment.py
ayush1298: added this

scripts/create_pr_results_comment.py
Samoed: To have better separation

.github/workflows/model-results-comparison.yaml
Samoed: I can't find such methods in `gh pr comment`

ayush1298: I am saying we could add them to solve this issue.

Samoed: We're limited by `gh pr` interface

ayush1298: Then what will be solution here? Maybe then we can skip comment, and only keep tests, though it will be difficult in terms of visual perspective

Samoed: I think we can keep only tests for now

ayush1298: Okay. Will remove the code for generating PR comment then.

scripts/create_pr_results_comment.py
ayush1298: @Samoed Should I keep all code in test file then only, as we are not generating PR comment?

ayush1298:
Remove PR comment generation part, and changed it to run only test

scripts/create_pr_results_comment.py
```diff
@@ -3,34 +3,37 @@
 
 Usage:
     gh pr checkout {pr-number}
-    python scripts/create_pr_results_comment.py [--models MODEL1 MODEL2 ...] [--output OUTPUT_FILE]
+    python scripts/create_pr_results_comment.py [--models MODEL1 MODEL2 ...] [--output-comparison FILE] [--output-diff FILE]
```
Samoed: Seems that you didn't update cli
```suggestion
    python scripts/create_pr_results_comment.py [--models MODEL1 MODEL2 ...] [--output-comparison FILE]
```

scripts/create_pr_results_comment.py
```diff
@@ -63,6 +69,124 @@ def get_diff_from_main() -> list[str]:
...
+        # Get the main_score from the first split/subset
+        for split_scores in filtered_result.scores.values():
+            for subset_score in split_scores:
+                main_score = subset_score.get("main_score")
+                if main_score is not None and not pd.isna(main_score):
+                    value = float(main_score)
+                    # Normalize percentage scores to decimal
+                    if value > 1:
+                        value /= 100
+                    return value
```
Samoed: Why you take only one value from full results?

scripts/create_pr_results_comment.py
```diff
@@ -63,6 +69,124 @@ def get_diff_from_main() -> list[str]:
...
+                old_revision = "unknown"
+
+        try:
+            with path.open("r") as f:
+                new_json = json.load(f)
+        except (json.JSONDecodeError, IOError):
+            continue
+
+        old_value = extract_main_score(old_json)
+        new_value = extract_main_score(new_json)
```
Samoed: You need also to filter old/new existing results by splits/subsets

scripts/create_pr_results_comment.py
ayush1298: I think I missed this one. Will correct it.

scripts/create_pr_results_comment.py
ayush1298: So, we want to compare each main_score across splits/subsets, or we want to aggregate?

Samoed: Aggregate, but on same splits/subsets

ayush1298: Is it correct that there will be 1st split, and then in each split there will be a subset?
So, here we should aggregate results of all subset in 1 split and then compare, or we should just compare 1-by-1, each subset in a split?

Samoed: > Is it correct that there will be 1st split, and then in each split there will be a subset?

Yes

> So, here we should aggregate results of all subset in 1 split and then compare, or we should just compare 1-by-1, each subset in a split?

I think would be better to compare 1-by-1 for now

ayush1298: Added these

tests/test_results_diff.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+import sys
+from pathlib import Path
+
+import pytest
+
+# Add scripts directory to path to import create_pr_results_comment
+scripts_dir = Path(__file__).parents[1] / "scripts"
+sys.path.insert(0, str(scripts_dir))
+
```
KennethEnevoldsen, line 20: Hmm not sure we want to import it like this - but I am not sure what the best approach is here (@Samoed any ideas?)

tests/test_results_diff.py
```diff
@@ -0,0 +1,64 @@
...
+
+def test_result_diffs_within_threshold():
+    """
+    Fail if any main_score delta exceeds configured thresholds.
+    
+    Environment variables:
+    - MAX_ALLOWED_PCT_CHANGE: Maximum percent change allowed (default: 0.05 = 5%)
+    """
+    max_pct_change = float(os.getenv("MAX_ALLOWED_PCT_CHANGE", "0.05"))
```
KennethEnevoldsen: typically we use an epsilon of 0.001 (so 0.5032 and 0.5039 are the considered the same score). I think we should use this here.

I think we should implement this as a variable in `mteb` and import it here.

```
from mteb import MTEB_SCORE_EPSILON
```

ayush1298: Can we define it here only? Why define in MTEB and import?

KennethEnevoldsen: We can define it here for now

ayush1298: Added this now.

tests/test_results_diff.py
Samoed: I think we can keep like this. I don't see better solutions

KennethEnevoldsen (commented):
Alright looks good - @ayush1298 can I ask you to create an example where it fails just to see that does as expect (then we can delete it again and merge)

ayush1298:
> Alright looks good - @ayush1298 can I ask you to create an example where it fails just to see that does as expect (then we can delete it again and merge)

This compares branch with main, so If I add example here, it will just skip. But, I have tested it locally, by creating new branch from these branch and modifying results there.

KennethEnevoldsen:
Not sure I get it - shouldn't it fail if you add an faulty example here?

ayush1298:
> Not sure I get it - shouldn't it fail if you add an faulty example here?

sorry, my bad. Actually I was mistakenly having old implementation in mind, which needs changing branch.
Added changed results for a model, and test is failing.

tests/test_results_diff.py
```diff
@@ -0,0 +1,61 @@
...
+    for _, row in diff_table.iterrows():
+        delta = abs(row["delta"])
+        
+        model_task = f"{row['model_name']}/{row['task_name']}"
+        
+        if delta > MTEB_SCORE_EPSILON:
+            violations.append(
+                f"  {model_task}: delta {delta:.4f} exceeds threshold {MTEB_SCORE_EPSILON}"
+            )
+
```
KennethEnevoldsen, line 191: Let us improve the logging here so it is easier to locate the diff
```suggestion
            violations.append(
                f"  {model_task}: The difference between the current score ({}) and the previous ({})  exceeds threshold of {MTEB_SCORE_EPSILON}"
            )
    
```

KennethEnevoldsen:
then of course we need to remove the file and then I think we are good

tests/test_results_diff.py
ayush1298: Changed these one.

ayush1298:
> then of course we need to remove the file and then I think we are good

Test is failing, and now the log is:

```
FAILED tests/test_results_diff.py::test_result_diffs_within_threshold - AssertionError: Main score changes exceed configured threshold (MTEB_SCORE_EPSILON=0.001):
    aari1995/German_Semantic_STS_V2/AmazonCounterfactualClassification: The difference between the current score (0.675315) and the previous (0.665315) exceeds threshold of 0.001
assert not ['  aari1995/German_Semantic_STS_V2/AmazonCounterfactualClassification: The difference between the current score (0.675315) and the previous (0.665315) exceeds threshold of 0.001']
```


I have removed the file

Samoed:
@ayush1298 Can you submit updated md file example?

ayush1298:
> @ayush1298 Can you submit an updated MD file example?

Which one? We are not generating that comment anymore.

tests/test_results_diff.py
```diff
@@ -0,0 +1,61 @@
...
+# Add scripts directory to path to import create_pr_results_comment
+scripts_dir = Path(__file__).parents[1] / "scripts"
+sys.path.insert(0, str(scripts_dir))
+
+from create_pr_results_comment import (
+    get_base_ref,
+    get_diff_from_main,
+    create_old_new_diff_table,
+)
```
Samoed: If we don't generate comment anymore then we can we move these functions inside test?

results/aari1995__German_Semantic_STS_V2/22912542b0ec7a7ef369837e28ffe6352a27afc9/AmazonCounterfactualClassification.json
Samoed, line 1: Can you rollback this deletion?

tests/test_results_diff.py
ayush1298: `get_diff_from_main` is used in this file also. So, will keep that here. And move other 2 functions in test file

results/aari1995__German_Semantic_STS_V2/22912542b0ec7a7ef369837e28ffe6352a27afc9/AmazonCounterfactualClassification.json
ayush1298: Sorry, will do it.

tests/test_results_diff.py
Samoed: But this is very small function and I think it's simpler to add new one in tests

ayush1298: Okay, will do it. This is to just avoid that import from script, right?

results/aari1995__German_Semantic_STS_V2/22912542b0ec7a7ef369837e28ffe6352a27afc9/AmazonCounterfactualClassification.json
```diff
@@ -82,7 +82,7 @@
             "ap_weighted": 0.140298
           }
         ],
-        "main_score": 0.665315,
```
Samoed, line 85: Can you also rollback this?

tests/test_results_diff.py
Samoed: Yes

results/aari1995__German_Semantic_STS_V2/22912542b0ec7a7ef369837e28ffe6352a27afc9/AmazonCounterfactualClassification.json
ayush1298: Yes, just kept it untill I moved all functions and checked its working correctly. Then will remove it

ayush1298:
Move all functions in test file, So
Test is failing as expected with error: 

```
E       AssertionError: Main score changes exceed configured threshold (MTEB_SCORE_EPSILON=0.001):
E           aari1995/German_Semantic_STS_V2/AmazonCounterfactualClassification: The difference between the current score (0.675315) and the previous (0.665315) exceeds threshold of 0.001
E       assert not ['  aari1995/German_Semantic_STS_V2/AmazonCounterfactualClassification: The difference between the current score (0.675315) and the previous (0.665315) exceeds threshold of 0.001']
```

I am rollbacking changed results file

KennethEnevoldsen:
Wonderful!

PR #445, opened 2026-03-14
Keep only one commit in cached branch

Samoed:
Now cached branch have a lot of commits which leads to very long clone (currently almost 3gb in CI). I've changed to store only 1 commit


isaac-chung:
Then can we revert to a previous commit if someone screws up? If not, then we should not do this.

Samoed:
> Then can we revert to a previous commit if someone screws up? 

This is just cache. I don't think this is a problem

isaac-chung (approved):

PR #465, opened 2026-03-31
F2LLM-v2 results on Thai and Spanish benchmark

Geralt-Targaryen:
### Checklist

As discussed in https://github.com/embeddings-benchmark/mteb/issues/3237 and https://github.com/embeddings-benchmark/mteb/issues/4321, I evaluated the F2LLM-v2 models on the Thai and Spanish benchmarks.

`MKQARetrieval`, `WisesightSentimentClassification.v2`, and `SpanishSentimentClassification.v2` are evaluated using these prompts:
```json
{
    "MKQARetrieval": "Given a question, retrieve the answer.",
    "WisesightSentimentClassification.v2": "Classify the sentiment of the given text.",
    "SpanishSentimentClassification.v2": "Classify the sentiment of the given text."
}
```
Other tasks are evaluated using the prompts in the current implementation.

I have two related questions:

1. Do I need to create a PR on model implementation to add relevant prompts each time new tasks are evaluated? Or does it suffice to show the prompts here?
2. For certain tasks that have existing results on other languages (e.g., MIRACLReranking), when I evaluated the models on the Thai benchmark, the existing results were erased. But then I evaluated the models on the Spanish benchmark, and the Thai results were not erased. I presume this is probably because different MTEB versions were used (existing results use v2.6.7; now I'm using v2.12.4). Currently I solved this issue by manually merging the new results into the existing file and setting the mteb_version field to the latter one. However, I think it would be better to automatically resolve this conflict.


Samoed:
Can you add prompts to your model meta?

Geralt-Targaryen:
I've updated the model meta. Do I need to created a PR to add the new prompts to model implementation as well?

Samoed:
> Do I need to created a PR to add the new prompts to model implementation as well?

Yes

KennethEnevoldsen:
> For certain tasks that have existing results on other languages (e.g., MIRACLReranking), when I evaluated the models on the Thai benchmark, the existing results were erased. But then I evaluated the models on the Spanish benchmark, and the Thai results were not erased. I presume this is probably because different MTEB versions were used (existing results use v2.6.7; now I'm using v2.12.4). Currently I solved this issue by manually merging the new results into the existing file and setting the mteb_version field to the latter one. However, I think it would be better to automatically resolve this conflict.

You should not automatically merge these - they were run using different MTEB versions and therefore shouldn't be merged (it makes reproducing the results potentially impossible).

A solution right now it rerun the full set with your current version or downgrading the version.

I do however, agree that this is not ideal, we would need to move the MTEB version to the individual subsets to avoid this issue. I would be more than happy to see an issue on this to outline the problem

Geralt-Targaryen:
I've created an issue. I'll rerun the models on all subsets.

The prompts have been added to the implementation btw (https://github.com/embeddings-benchmark/mteb/pull/4336)

Geralt-Targaryen:
Rerunning the full set would take a really long time for some tasks. So I've rerun the Thai and Spanish subsets using the older version instead. The results are identical anyway. @KennethEnevoldsen

KennethEnevoldsen (approved):

PR #478, opened 2026-04-05
fix: set main score to LRAP instead of accuracy for multilabel clf audio tasks

KennethEnevoldsen:
run using:
```
import json
from pathlib import Path

results = Path(__file__).parent / "results"
# get all json files in results
json_files = results.glob("**/*.json")

tasks = {"FSD2019Kaggle", "FSD50K", "BirdSet"}

for f in json_files:
    if f.stem in tasks:
        with f.open() as fp:
            data = json.load(fp)

        for split, scores in data["scores"].items():
            for i, score in enumerate(scores):
                data["scores"][split][i]["main_score"] = score["lrap"]
        with f.open("w") as fp:
            json.dump(data, fp, indent=2)
```

Note test will fail due to a large diff in main score, that is expected

### DIFF
--- a/results/LCO-Embedding__LCO-Embedding-Omni-3B/eea763cfaf673e955ae86c64968896a3fea70189/BirdSet.json
+++ b/results/LCO-Embedding__LCO-Embedding-Omni-3B/eea763cfaf673e955ae86c64968896a3fea70189/BirdSet.json
@@ -71,7 +71,7 @@
         "lrap": 0.270639,
         "f1": 0.037509,
         "hamming": 0.036672,
-        "main_score": 0.030503,
+        "main_score": 0.270639,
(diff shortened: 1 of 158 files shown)


isaac-chung (approved):

AdnanElAssadi56 (approved):

AdnanElAssadi56:
@isaac-chung Do we need to bypass test before merging?

isaac-chung:
We're good. Tests fail because the score diff was too high, but that's expected when we switch main metrics.

isaac-chung:
I don't have the rights to bypass. @KennethEnevoldsen might. Or we could comment out the test here. And add it back in a new PR.

KennethEnevoldsen:
Sorry let me bypass those

PR #506, opened 2026-05-01
Parquet cache

nv-alaiacano:
This PR creates a parquet formatted version of the BenchmarkResults dataset. It reduces the size of the results file from 98MB (_compressed_ json) to 65MB of parquet.

It's a preparation for optimizations in the visualizer in the `mteb` repository, where we will be able to more efficiently load and aggregate results data.

### Checklist - n/a this is not a model result submission.

- [ ] The results submitted are obtained using the reference implementation


.github/workflows/generate_cached_results.yml
Samoed, line 1: I'm not sure if we want to update `generate_cached_results.yml` with this. I think would be better to create separate action for this and push cache to HF repo

scripts/generate_cached_results.py
Samoed, line 1: Can we also create separate script for parquet creation?

nv-alaiacano: Yeah, i can break it out into a separate script with some shared functionality between them for loading the BenchmarkResults

.github/workflows/generate_cached_results.yml
nv-alaiacano: The existing action pushes the zipped json file to the [cached-data](https://github.com/embeddings-benchmark/results/tree/cached-data) branch in this repo, which is later read in the [mteb ResultCache](https://github.com/embeddings-benchmark/mteb/blob/main/mteb/cache.py#L324-L333). It seems like this is intentional so that the large data files _aren't_ sent to HF. Please correct me if I'm wrong!

I will break this out into a separate workflow but it might be a little delicate, because there will be a race condition of two workflows both force-pushing to the same branch, and each will have to be configured to not modify the other file(s). I'll make a reusable action to do the push to branch carefully - hopefully this won't be too much extra complexity. An alternative would be setting up a different orphaned cache branch to push to.

Samoed: > It seems like this is intentional so that the large data files aren't sent to HF.

Yes, we are currently saving it like this because it was easy to integrate and JSON doesn't have good visualization on HF, but if we save as parquet it would be easy to use from HF and easy to use for others.
> An alternative would be setting up a different orphaned cache branch to push to.

Yes, this would be easier, but I think it would be even easier to work with and store results on HF Hub.

scripts/generate_cached_results.py
```diff
@@ -30,6 +37,70 @@
...
+        ("date", pa.timestamp("us", tz="UTC")),
+        ("split", pa.string()),
+        ("hf_subset", pa.string()),
+        ("languages", pa.list_(pa.string())),
+        ("main_score", pa.float64()),
+    ]
+)
+
+
+def _benchmark_results_to_table(all_results: BenchmarkResults) -> pa.Table:
```
KennethEnevoldsen, line 64: wouldn't we rather want this on the `BenchmarkResults` object as `to_parquet` (and `from_parquet`)?

KennethEnevoldsen (commented):
Another option is also to move it to `mteb,` and include it along with the Docker container (@nv-alaiacano, I believe you already mentioned this). 

It would require us to download this repo in the CI when creating the Docker container. It would however simplify the CI workflows a bit

KennethEnevoldsen:
Hi @nv-alaiacano, any update on this?

Samoed:
I created a draft PR https://github.com/embeddings-benchmark/mteb/pull/4716 with creating parquet export, and it's usage in leaderboard, but WIP

PR #512, opened 2026-05-03
Qwen2.5 and Qwen3 omni results on MAEB

isaac-chung:
- Qwen/Qwen2.5-Omni-3B
- Qwen/Qwen2.5-Omni-7B
- Qwen/Qwen3-Omni-30B-A3B-Instruct
- Qwen/Qwen3-Omni-30B-A3B-Thinking

### Checklist


KennethEnevoldsen:
@isaac-chung edited the comment to not fill out the " I solemnly swear..." field since you are not the model author.

PR #562, opened 2026-06-09
Removed is_public column and exclude In Training Data column from being bold in PR comment

### LINKED ISSUE #4782: `is_public` is always bold when submitting results
### Description of the feature

When submitting a results we use bold to detect problematic values, but since `is_public` was added it is always bold even though it is just metadata. The solution is to determine bold to exclude metadata columns like "is_public".

Image

seen in https://github.com/embeddings-benchmark/results/pull/561

ayush1298:
Close https://github.com/embeddings-benchmark/mteb/issues/4782


Samoed:
Can you completely remove is_public column?

scripts/create_pr_results_comment.py
```diff
@@ -211,7 +211,7 @@ def create_comparison_table(
 
 
 def highlight_max_bold(
-    df: pd.DataFrame, exclude_cols: list[str] = ["task_name"]
+    df: pd.DataFrame, exclude_cols: list[str] = ["task_name", "is_public"]
```
Samoed: Can you add here also `In Training Data`?

ayush1298: I have thought of that, but don't we have to flag that column?

ayush1298:
> Can you completely remove is_public column?

isn't it required?

Samoed:
No, we don't need this column for comments in the results repo. This column was added for results aggregation on the leaderboard.

ayush1298:
> No, we don't need this column for comments in the results repo. This column was added for results aggregation on the leaderboard.

okay, will remove it completely then. And should I add `In Training Data` to exclude_cols?

Samoed:
Yes, `In Training Data` shouldn't be highlighted.

Samoed:
Seems correct now. Can you rollback changes in scores and we can merge then

ayush1298:
> Seems correct now. Can you rollback changes in scores and we can merge then

done

Samoed (approved):

PR #563, opened 2026-06-09
fix results table truncation issue

### LINKED ISSUE #4110: Result tables getting truncated due to GitHub API limitations
> Note: Content truncated due to GitHub API limits. See the full report in the workflow artifacts.

Also mentioned here: https://github.com/embeddings-benchmark/results/pull/418#issuecomment-3926533828

I would suggest simplifying the message a bit.

Here is a few suggestion:

**remove list of tasks as it is already in the table:** 

> Tasks: AmazonCounterfactualClassification, ArXivHierarchicalClusteringP2P, ArXivHierarchicalClusteringS2S, ArguAna, AskUbuntuDupQuestions, BIOSSES, Banking77Classification, BiorxivClusteringP2P.v2, CQADupstackGamingRetrieval, CQADupstackUnixRetrieval, ClimateFEVERHardNegatives, FEVERHardNegatives, FiQA2018, HotpotQAHardNegatives, ImdbClassification, MTOPDomainClassification, MassiveIntentClassification, MassiveScenarioClassification, MedrxivClusteringP2P.v2, MedrxivClusteringS2S.v2, MindSmallReranking, SCIDOCS, SICK-R, STS12, STS13, STS14, STS15, STS17, STS22.v2, STSBenchmark, SprintDuplicateQuestions, StackExchangeClustering.v2, StackExchangeClusteringP2P.v2, SummEvalSummarization.v2, TRECCOVID, Touche2020Retrieval.v3, ToxicConversationsClassification, TweetSentimentExtractionClassification, TwentyNewsgroupsClustering.v2, TwitterSemEval2015, TwitterURLCorpus

**Remove google/gemini from table**

It was mostly there to show the best score, but we also have the "max score" so we could remove that (I would like to keep it though)

**Reduce decimals**

Currently, we have e.g. `0.6492`, we could probably make it `.649`

**Remove training data list as it has since been added to table**

**Remove orgs from model names**

probably not needed

**Prioritize results where the model performs higher that max**

If it will be shortened this might be great to have at the top

ayush1298:
Closes https://github.com/embeddings-benchmark/mteb/issues/4110

Changes https://github.com/embeddings-benchmark/results/blob/main/.github/workflows/model-results-comparison.yaml#L79-L89 to add the message only when limit exceeded.

Also removed repeated things like: 
1. Removed list of Tasks as it's already in Table
2. Remove the training datasets list below the table, as its already added
3. Added formatting to convert floats to 3 decimal places and remove leading zeroes e.g. 0.6492 to .649


ayush1298:
@Samoed Somehow even after the check:
`Note: Content truncated due to GitHub API limits. See the full report in the workflow artifacts.`
these line is there in comment, though there are very little changes

.github/workflows/model-results-comparison.yaml
```diff
@@ -81,9 +81,15 @@ jobs:
...
-        head -c 65300 model-comparison.md > model-comparison-truncated.md
-        echo -e "\n\n---\n**Note:** Content truncated due to GitHub API limits. See the full report in the workflow artifacts." >> model-comparison-truncated.md
-        mv model-comparison-truncated.md model-comparison.md
-
-        gh pr comment ${{ steps.pr_info.outputs.pr_number }} --body-file model-comparison.md --create-if-none --edit-last
+        if [ -f model-comparison.md ]; then
+          FILE_SIZE=$(wc -c < model-comparison.md | tr -d ' ')
+          if [ "$FILE_SIZE" -gt 65300 ]; then
+            echo "Truncating model-comparison.md to fit GitHub API limit..."
+            head -c 65300 model-comparison.md > model-comparison-truncated.md
```
Samoed, line 88: Shouldn't this be then? 
```suggestion
            head -c 65299 model-comparison.md > model-comparison-truncated.md
```

scripts/create_pr_results_comment.py
```diff
@@ -217,13 +217,19 @@ def highlight_max_bold(
 ) -> pd.DataFrame:
     result_df = df.copy()
 
+    def format_score(x):
```
Samoed, line 220: I don't think that this will save space

ayush1298: This was suggested by Kenneth in issue description. Though, I guess in terms of formatting, its better

.github/workflows/model-results-comparison.yaml
ayush1298: Actual limit is 65536 characters, should I set that, I try to leave some buffer

Samoed:
> @Samoed Somehow even after the check: `Note: Content truncated due to GitHub API limits. See the full report in the workflow artifacts.` these line is there in comment, though there are very little changes

Actions with `pull_request_target` run action from main. You can see in the action log that it's running the previous version of your action (it's running from `main`), but script is running from your branch

Samoed:
Formatting looks a bit strange. Some numbers have 0, while others don't

KennethEnevoldsen:
I think the biggest issue is that:

> "Note: Content truncated due to GitHub API limits. See the full report in the workflow artifacts."

Which makes it seem like the message was truncated even though it wasn't

Samoed:
> I think the biggest issue is that:
> 
> > "Note: Content truncated due to GitHub API limits. See the full report in the workflow artifacts."
> 
> Which makes it seem like the message was truncated even though it wasn't

This will be resolved after merge. Currently action's yaml is taken from main branch due to `pull_request_target`

ayush1298:
> Formatting looks a bit strange. Some numbers have 0, while others don't

I have fixed it and its now working correctly. You can see the updated comment of Github Bot

Samoed (approved):

ayush1298:
I have reverted scores

Samoed (approved):

PR #627, opened 2026-07-19
MVEB(beta) video results: EximiusLabs/fusion-embedding-2-2b-preview

abtonmoy:
Adds MVEB(beta) video results for `EximiusLabs/fusion-embedding-2-2b-preview` (generation 2 of the Fusion Embedding family; model merged in embeddings-benchmark/mteb#4952, MAEB audio results in #625). 17 video tasks, `mteb==2.18.2`, revision `2d30494e4ef5`, dropped into the model's existing revision directory alongside its audio results so both benchmarks share one model card.

Matched to the generation-1 video suite (#624). The two generations score identically within evaluation noise, which is expected by construction: generation 2's deep adapters are gated to audio inputs, so on video the gate is closed and the video path is the frozen Qwen3-VL-Embedding-2B base, byte-for-byte the same as generation 1.

| Task | gen-2 | gen-1 |
|---|---:|---:|
| HumanAnimalCartoonVPairClassification | .880 | .880 |
| MSVDT2VRetrieval | .824 | .826 |
| UCF101VideoZeroShotClassification | .811 | .813 |
| MSVDV2TRetrieval | .777 | .779 |
| VATEXT2VRetrieval | .749 | .756 |
| HMDB51Classification | .728 | .728 |
| WorldSenseVideoClassification | .680 | .680 |
| BreakfastClassification | .679 | .679 |
| DiDeMoV2TRetrieval | .591 | .595 |
| VALOR32KT2VRetrieval | .557 | .561 |
| AVMemeVideoClassification | .480 | .480 |
| WorldSenseVideoZeroShot | .412 | .411 |
| AVMemeExamT2VRetrieval | .358 | .363 |
| AudioCapsAVT2VRetrieval | .271 | .272 |
| MELDVideoZeroShot | .218 | .226 |
| RAVDESSVideoClustering | .150 | .150 |
| MELDVideoClassification | .146 | .146 |

`AudioCapsAVT2VRetrieval` uses AudioCaps (declared in `training_datasets`) and shows as in-training; all other tasks are outside the training corpus.

Validation: `tests/test_correct_folder_structure.py` and `tests/test_ensure_correct_metadata.py` pass locally against the full tree (revision field matches the folder name).

### Checklist

- [x] My model has a model sheet, report, or similar
[...]

PR #634, opened 2026-07-24
Add fusion-embedding-2-2b-preview v0.3-preview MAEB(audio-only) results

abtonmoy:
Adds MAEB(audio-only) results for `EximiusLabs/fusion-embedding-2-2b-preview` revision `v0.3-preview` (commit `8aa29168f03d043fe4246a1f40ca10a0ed0595d8`), 19 tasks, under the model's revision directory. This is the retrieval-tuned flagship of our v0.3 release.

- Weights and model card: https://huggingface.co/EximiusLabs/fusion-embedding-2-2b-preview (revision `v0.3-preview`)
- Technical report: https://arxiv.org/abs/2607.18666

**Training-data disclosure (re: the final checklist item).** One reported task, `CommonLanguageAgeDetection`, is derived from Common Voice, which is part of this model's training data. We report it for completeness; its score (0.171) is below the field median and does not inflate the reported mean. No other reported task overlaps the training data. In particular, these checkpoints are not trained on CREMA-D, so the `CREMA_D`, `CREMA_DClustering`, and `CREMADPairClassification` cells are clean.

### Checklist


Samoed:
Revision of results should be `9451b840f0d1d95440f3836e9a7f600a833d663f`. Your revision also should be in main, not in different branch

abtonmoy:
Thanks for the review, and good catch on the revision.

One clarification: `9451b840f0d1d95440f3836e9a7f600a833d663f` is a superseded earlier revision of this repo. It predates the current release (its `config.json` has `version: null`) and has different weights, so the results here were not produced by it. These results are from our current model, `v0.3-preview`.

I've re-filed both PRs under `1720d8b16af578d794d7b21ee7b829281941899d`, which is the `v0.3-preview` flagship commit on the `main` branch. It has the same `model.safetensors` as the `v0.3-preview` tag (I had originally filed under the tag commit, which is what put the revision off `main`). `model_meta.json` is updated to match, and both #634 and #635 now point at this on-main revision.

The older `9451b840` and `2d30494` result directories are from that superseded revision; I'm happy to open a cleanup PR to remove them if you'd prefer everything consolidated on the current revision. Let me know if the model registration should be updated to `1720d8b1...` as well.

Samoed:
> The older 9451b840 and 2d30494 result directories are from that superseded revision

You need to update revision in mteb repo then

abtonmoy:
Makes sense. Opened embeddings-benchmark/mteb#5028 to update the registration for `EximiusLabs/fusion-embedding-2-2b-preview` from `9451b840…` to `1720d8b16af578d794d7b21ee7b829281941899d` (the v0.3-preview flagship on `main`), so it matches the results here and in #635.

I'll also open a cleanup PR to remove the superseded `9451b840` and `2d30494` result directories so everything consolidates on the current revision. Thanks for the review.

abtonmoy:
Correction on the cleanup I mentioned: those two directories aren't stale duplicates. `9451b840…` holds the v0.2 MAEB(beta) audio results, and `2d30494…` holds the merged MVEB video results (#627). #634/#635 re-file the audio under `1720d8b1`, but there's no v0.3 video submission yet, so removing `2d30494…` would drop the model's video results entirely.

So I'll leave both directories in place rather than delete anything, and consolidate the video results onto `1720d8b1` in a follow-up once they're re-run. Happy to do it differently if you'd prefer.

Samoed:
Overall what's differenece between 0.2 and 0.3? Does video part was modified?

abtonmoy:
Good question — the difference between v0.2 and v0.3 is entirely on the audio path.

fusion-embedding-2 is a frozen VLM base with modality-gated adapters that fire only on audio. v0.3 takes v0.2 and adds an AudioCaps fine-tune on those audio adapters — that's what moves the caption-retrieval tasks it targets (e.g. ClothoT2ARetrieval `.278 → .300`), with some shift on the other zero-shot audio tasks.

The video path was not modified. Because the adapters are audio-gated, the image/video/text forward passes are bit-for-bit identical to the frozen base and unchanged across v0.2 and v0.3 — we verify byte-equality on those modalities after every training run. So the MVEB video results (#627) hold exactly for v0.3; re-running them on `1720d8b1` reproduces the same numbers.

If it helps keep revisions consistent, I'm happy to re-file the video results under `1720d8b1` (the on-`main` v0.3 revision now registered in mteb) so audio and video sit on the same revision. The numbers don't change — only the revision the directory is filed under.

abtonmoy:
Hi @Samoed, following up on this one. The answer above covers the full v0.2 to v0.3 delta: the fine-tune touches only the audio adapters, so the video and text paths are unchanged and the merged MVEB and MTEB results still apply as-is.

One thing that changed since we last spoke: the ModelMeta revision was bumped to v0.3-preview in embeddings-benchmark/mteb#5028, so the model the leaderboard points at now matches these results. #635 completes the remaining tasks for the 30-task board on that same pinned revision, so reviewing the two together might be easiest.

Happy to rerun anything or split the PRs differently if that helps.

Samoed:
Hi! I rerun tests, can you remove duplicated results? 
```
EximiusLabs/fusion-embedding-2-2b-preview:
  Task: BeijingOpera
    Revisions present: ['1720d8b16af578d794d7b21ee7b829281941899d', '9451b840f0d1d95440f3836e9a7f600a833d663f']
    Retained revision: 1720d8b16af578d794d7b21ee7b829281941899d
    Should delete from revisions: ['9451b840f0d1d95440f3836e9a7f600a833d663f']
  Task: GTZANAudioReranking
    Revisions present: ['1720d8b16af578d794d7b21ee7b829281941899d', '9451b840f0d1d95440f3836e9a7f600a833d663f']
    Retained revision: 1720d8b16af578d794d7b21ee7b829281941899d
    Should delete from revisions: ['9451b840f0d1d95440f3836e9a7f600a833d663f']
  Task: GTZANGenre
    Revisions present: ['1720d8b16af578d794d7b21ee7b829281941899d', '9451b840f0d1d95440f3836e9a7f600a833d663f']
    Retained revision: 1720d8b16af578d794d7b21ee7b829281941899d
    Should delete from revisions: ['9451b840f0d1d95440f3836e9a7f600a833d663f']
  Task: VehicleSoundClustering
    Revisions present: ['1720d8b16af578d794d7b21ee7b829281941899d', '9451b840f0d1d95440f3836e9a7f600a833d663f']
    Retained revision: 1720d8b16af578d794d7b21ee7b829281941899d
    Should delete from revisions: ['9451b840f0d1d95440f3836e9a7f600a833d663f']
```

abtonmoy:
Done, removed the four duplicated v0.2-revision results (BeijingOpera, GTZANAudioReranking, GTZANGenre, VehicleSoundClustering under 9451b840); each task now resolves to the retained v0.3 revision only. Dedup should come back clean on a rerun.

tomaarsen:
Congratulations on the integration @abtonmoy. Out of curiosity, have you also considered an integration with Sentence Transformers? I think it might be pretty well suited, e.g. akin to https://huggingface.co/LCO-Embedding/LCO-Embedding-Omni-3B-2605. 

- Tom Aarsen

abtonmoy:
Hi Tom, thanks, and yes, very interested. Sentence Transformers is the natural home for this model, and the v5.4 multimodal `encode()` API (text, image, audio, video, and multimodal dicts) matches what the model does one to one.

One structural question before I build it. LCO-Embedding-Omni loads natively because its backbone is in stock transformers, so plain ST config files are enough. Ours is not: the encoder adds modality-gated adapters and an audio frontend that live in our `fusion-embedding` package, which is why our vLLM and SGLang integrations ship as plugins. For ST, is the recommended route custom modules on the HF repo with `trust_remote_code`, or would you rather see the architecture land in transformers first and ST load it natively? Happy to do either.

tomaarsen:
Glad to hear it! For your question: the former is the recommended route nowadays. Transformers cannot justify adding every novel architecture, so usually model authors start with custom code and then a transformers integration could potentially flow from that eventually. Then initially it would require `trust_remote_code=True` in both transformers and Sentence Transformers, but in my experience this is not much of a limiting factor for usage, especially as users can get extra safety by pinning a `revision` when loading.

- Tom Aarsen

abtonmoy:
It's live: `SentenceTransformer("EximiusLabs/fusion-embedding-2-2b-preview", trust_remote_code=True)`. Single `InputModule` wrapping our `fusion-embedding` package, so text, image, audio and video go through the same path as the native loader and come out bitwise identical.

One thing worth a line in the custom-modules docs: `get_imports` skips `ast.Try` blocks but otherwise treats every import as load-time required, including inside a function body. A bare `from PIL import Image` in an image helper made Pillow mandatory just to construct the model, with the error advising `pip install PIL`. Easy to hit in any custom module with an optional dependency.

Happy to send a short docs PR for that, and for the multimodal version floor if it is worth documenting.

tomaarsen:
Oh, that's a good quirk to know about, I didn't realise. I'm happy to accept a PR for it on the Sentence Transformers repository!

- Tom Aarsen

PR #663, opened 2026-08-04
Add missing nomic moe results for RTEB, MMTEB and English MTEB

KennethEnevoldsen:
Saw it was missing when looking into MoE performance.

### Checklist


results/nomic-ai__nomic-embed-text-v2-moe/1066b6599d099fbb93dfcb64f9c37a7c9e503e85/run_settings.jsonl
KennethEnevoldsen, line 1: @Samoed this one seem to have gotten too big and fails the file size check. 

I have chosen to add stuff to the check file size to reduce it by combining the subsets:

```
{"task":"MassiveIntentClassification","split":"test","version":{"mteb":"2.18.13","torch":"2.6.0+cu124","sentence-transformers":"4.0.2","flash-attn":null,"transformers":"4.57.6"},"encode_kwargs":{"batch_size":32},"subsets":["af","am","ar","az","bn","cy","da","de","el","en","es","fa","fi","fr","he","hi","hu","hy","id","is","it","ja","jv","ka","km","kn","ko","lv","ml","mn","ms","my","nb","nl","pl","pt","ro","ru","sl","sq","sv","sw","ta","te","th","tl","tr","ur","vi","zh-CN","zh-TW"]}
```

but only for cases where there the args match. We could probably also move to the script that generated run settings just to keep it more compact. WDYT?

Also in this process fixed a bug in the flores script that always failed (but was caught).

KennethEnevoldsen: cc @ayush1298 as well

Samoed: Yes, I think tihs is good idea. We also can combine splits

reduce_large_json_files.py
```diff
@@ -1,43 +1,75 @@
...
+            continue
+        print(f"Resizing {file} to have no indentations")
+        with file.open("r") as f:
+            data = json.load(f)
+
+        with file.open("w") as f:
+            json.dump(data, f, indent=None)
+
+
+def collapse_run_settings():
```
Samoed, line 41: Maybe better to this in main code rather than in test?

PR #664, opened 2026-08-04
Add CC0 license

KennethEnevoldsen:
I assume CC0 is fine? We could also do MIT, but it is unclear to me to what extent this is data.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,121 @@
+Creative Commons Legal Code
+
+CC0 1.0 Universal
+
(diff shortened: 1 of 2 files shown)


Samoed (approved):
I think CC0 is fine. Probably we also need to add to HF repo

Muennighoff (approved):

Samoed:
I added license to hf repo https://huggingface.co/datasets/mteb/results/commit/1a8d558fec31b8d1afae46957a0c73363462ac4c

PR #693, opened 2026-08-28
Fix Querit/Querit-4B weblinx results

moshesbeta:
This PR completes the WebLINXCandidatesReranking results for Querit-4B.

All six splits were evaluated in parallel during the original evaluation run. Due to an issue in our result-merging script, only the validation and test_iid splits were included in the previously submitted JSON, while the results for test_cat, test_geo, test_vis, and test_web were inadvertently omitted.

We have recovered the original evaluation outputs for these four splits and included them in this PR. These results use the same MTEB version (2.14.8) and evaluation setup as the previously submitted WebLINX results. No model inference, re-evaluation, or changes to the model or evaluation setup were performed for this PR.

### Checklist


Samoed (approved):

terryqj0107:
Hi @Samoed 

It seems the restored result still records `mteb_version: 2.14.8`, and the PR notes that no re-evaluation was performed.

The WebLINX `mrr_at_10` tie-breaking bug documented in embeddings-benchmark/mteb#5092 was only fixed in MTEB 2.18.19. Therefore, adding the missing splits does not make these pre-fix MRR values comparable to results produced by the corrected evaluator.

The resulting MRR score is not merely outdated; it is artificially inflated by the known metric issue in MTEB 2.14.8. This is also visible from the internal inconsistency of the submitted metrics: for example, the restored result reports `mrr_at_10 = 0.306404`, while `ndcg_at_10 = 0.19317`, `map_at_10 = 0.14878`, `precision_at_10 = 0.0342`, and `accuracy = 0.08112`. These supporting metrics are substantially lower than those of several leaderboard models that Querit-4B now appears to outperform according to the WebLINX main score. The unusually high MRR is therefore not evidence of correspondingly strong ranking quality; it is consistent with the already confirmed tie-breaking bug inflating MRR under score ties.

I have added the details to the original issue:
https://github.com/embeddings-benchmark/mteb/issues/5092

Could this result be hidden from the current leaderboard until all six splits are recomputed with MTEB >= 2.18.19?

https://mteb-leaderboard.hf.space/benchmark/MTEB(Multilingual%2C%20v2)?types=Reranking&tab=perf_task
image

Samoed:
I'll reply in issue

PR #695, opened 2026-08-31
Add MIEB(Multilingual) results for dnotitia/DNA-VL-STEER-2B

dhk1349:
## Results for dnotitia/DNA-VL-STEER-2B

MIEB(Multilingual), **130/130 tasks**. Multilingual multimodal embedding model built on
`Qwen/Qwen3-VL-Embedding-2B` (2B params, 2048-dim), calibrated for cross-lingual alignment
across 36 languages.

**Model card:** https://huggingface.co/dnotitia/DNA-VL-STEER-2B
**Evaluated revision:** `521a9cead4b31920579f80dfc9609a7969ddba86`
**Reference implementation:** merged in embeddings-benchmark/mteb#5313, shipped in mteb 2.20.3
(`mteb.get_model_meta("dnotitia/DNA-VL-STEER-2B")` resolves on a clean install)

Run with mteb 2.20.2/2.20.3 through the registered `Qwen3VLEmbeddingWrapper` on H200 GPUs.

### MIEB(Multilingual) — mean per task type

| Task type | n | Score |
|---|---:|---:|
| DocumentUnderstanding | 10 | 84.00 |
| VisualSTS(eng) | 7 | 80.73 |
| ImageClustering | 5 | 78.88 |
| ImageClassification | 22 | 72.59 |
| VisionCentricQA | 6 | 71.37 |
| VisualSTS(multi) | 2 | 71.02 |
| Any2AnyMultilingualRetrieval | 3 | 70.99 |
| ZeroShotClassification | 23 | 57.55 |
| Compositionality | 7 | 44.94 |
| Any2AnyRetrieval | 45 | 43.20 |

### Checklist

- [x] My model has a reference implementation in `mteb/models/model_implementations/`
      (embeddings-benchmark/mteb#5313, merged)
- [x] My model is available publicly on Hugging Face
- [x] I *solemnly swear* that for all results submitted I have not trained on the evaluation
      dataset including training splits. If I have, I have disclosed it clearly.


KennethEnevoldsen (commented):
thanks for the PR @dhk1349, it seems like performance is surprisingly high on some tasks e.g. CVBenchDepth which is .902 vs previous max of .748. Same for CVBenchDistance. The rest are less noticeable. Can I ask you to double check for leakage on these datasets

dhk1349:
I truly appreciate for reviewing and thanks for taking a  careful look —  and I checked. The short version is that those
numbers come from the base checkpoint rather than from our training.

`DNA-VL-STEER-2B` is an alignment-calibrated version of
[`Qwen/Qwen3-VL-Embedding-2B`](https://huggingface.co/Qwen/Qwen3-VL-Embedding-2B).
I ran the unmodified base model on the same CV-Bench tasks:

| task | Qwen3-VL-Embedding-2B (base) | DNA-VL-STEER-2B | delta |
|---|---:|---:|---:|
| CVBenchDepth | 88.17 | 90.17 | +2.00 |
| CVBenchDistance | 68.50 | 71.00 | +2.50 |
| CVBenchRelation | 81.23 | 82.92 | +1.69 |

So the base model is already ~13 points above the previous best on Depth, and our
contribution is ~2 points, consistent across all three tasks.

Two things that I think account for the spread between models:

**Depth / Distance / Relation are binary** (2 choices — only Count has 4–6), so chance
is 50. About half the published entries sit at 44.7–53.3, i.e. at chance, and
essentially everything above ~57 has a VLM backbone. This looks like the architectural
split CV-Bench was designed to expose rather than an anomaly.

**Our training data doesn't reach these datasets.** The calibration corpus is a CC12M
subset paired with LLM-generated translations (36 languages). We train only the first
few transformer blocks and leave the vision encoder untouched. CV-Bench draws its
images from ADE20K / COCO / Omni3D and generates the questions from 3D annotations,
so there's no overlap with what we trained on. Also for the base model, we found no evidence that base model used the CVBench*(and other datasets)s' test set for training when reading the technical report of Qwen3-VL-Embedding. 

More generally, our scores track the base model closely almost everywhere — on the 23
tasks where I currently have both, 16 are within ±3 points (median +0.5). The one
large intended gain is multilingual retrieval (XM3600T2IRetrieval 59.82 → 70.90),
which is what the calibration targets. Relevant here: our CV-Bench deltas (+1.7 to
+2.5) sit right in the middle of that distribution — there's no task-specific jump on
the datasets you flagged.

On the base model's own pretraining data I genuinely can't make a claim — we don't
have visibility into it. But the base checkpoint is public and this reproduces in
under a minute if you'd like to check independently:

```python
import mteb
meta = mteb.get_model_meta("Qwen/Qwen3-VL-Embedding-2B")
mteb.evaluate(meta, mteb.get_task("CVBenchDepth"))
```

Happy to run the base model on the remaining CV-Bench tasks or anywhere else that
would help.

KennethEnevoldsen:
thanks for the clarification @dhk1349 -- I would love to see the base model runs, though I do think the evidence provided is enough to merge this. Feel free to submit the base models runs in a seperate PR

Congratulations on the release! :)

PR #718, opened 2026-09-15
Add AfriMTEB results for bge-m3, AfriE5-Large-instruct, multilingual-e5-large(-instruct), bm25s

O-Nicolinho:
…

These are the JSONs for around half of the models we evaluated on AfriMTEB, I'll create the PR for the other part as soon as I am able. The models included here are: bge-m3, AfriE5-Large-instruct, multilingual-e5-large, multilingual-e5-large-instruct, and bm25s. I am excluding SIB200Classification.v2 because I believe the v2 version does not currently run on MTEB, and it seems like the [v3 version has not been created](https://github.com/embeddings-benchmark/mteb/pull/5007). I am also excluding XM3600T2IRetrieval since text only models cannot run it. The new files are AfriMTEB tasks that have not had results uploaded to MTEB yet. The rewritten files simply overwrite the results uploaded before the (numerous) fixes to the implementation of the AfriMTEB benchmark were live on MTEB. 
### Checklist

- [X] My model has a model sheet, report, or similar
- [X] The results submitted are obtained using the reference implementation
- [X] My model is available, either as a publicly accessible API or publicly on e.g., Huggingface
- [X] I *solemnly swear* that for all results submitted I have not trained on the evaluation dataset including training splits. If I have, I have disclosed it clearly.


results/BAAI__bge-m3/5617a9f61b028005a4858fdac845db406aefb181/BibleNLPBitextMining.json
Samoed, line 1: Seems that you deleted some results for existing tasks. Can you update to keep them?

Samoed (changes requested):

results/BAAI__bge-m3/5617a9f61b028005a4858fdac845db406aefb181/BibleNLPBitextMining.json
O-Nicolinho: My apologies, yes you're right. It should be fixed now.

Samoed (approved):

PR #720, opened 2026-09-16
results: reference models for MTEB(ita, v1)

albertobarnabo:
Reference-model results for `MTEB(ita, v1)` (embeddings-benchmark/mteb#5479), on the tasks they were still missing, so the benchmark passes `test_reference_model_coverage`: `multilingual-e5-small`, `static-similarity-mrl-multilingual-v1`, `potion-multilingual-128M`, `F2LLM-v2-80M`, and `baseline-bm25s` on the retrieval and reranking tasks it is eligible for. `JuriFindITRetrieval` is already in via #719. Also `multilingual-e5-large` on the full benchmark, as a reference point for Italian.

Included on purpose: `XStance`, which is *not* in the benchmark. It is the evidence for leaving it out: every model lands below the 0.539 average precision of a random ranking.

Generated with `mteb.get_model` / `mteb.evaluate`, mteb 2.20.12, one A100. Only new files: 26 result JSONs, nothing existing modified or deleted. Where a task already had a result under an older revision of e5-large or bm25s, my copy is left out.

Drafted with Claude Code; I ran and verified every number.


results/intfloat__multilingual-e5-large/ab10c1a7f42e74530fe7ae5be82e6d4f11a719eb/MintakaRetrieval.json
Samoed, line 1: Seems you're deleting a lot of exsiting results with new submission

Samoed (changes requested):

Samoed (approved):

